PR #58, opened 2019-05-22
Issue #9 fixing error message for no internet

mehrler12:
#9 The "No Internet Connection" error messages are triggered by a specific error message returned from the Github API. This error message is different on windows and mac so I added a check for each. Given that there is no internet connection we aren't able to check http error codes. I also future proofed it slightly by only checking for the relevant part of the error message.

### DIFF
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -125,6 +125,7 @@ function getUserInfo(callback) {
 
   ghme.info(function(err, data, head) {
     if (err) {
+      console.log(err);
       if (err.toString().indexOf("OTP") !== -1)
       {
         github.auth.config({
@@ -139,7 +140,7 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
-      else if (err == "Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443"){
+      else if (err.toString().indexOf("ENOTFOUND") != 0 || err.toString().indexOf("ENOENT") != 0){
         displayModal("No internet connection");
       }else{
         displayModal(err);
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -139,7 +139,7 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
-      else if (err.toString().indexOf("ENOTFOUND") != 0 || err.toString().indexOf("ENOENT") != 0){
+      else if (err.errno == "ENOTFOUND" || err.errno =="ENOENT"){
         displayModal("No internet connection");
       }else{
         displayModal(err);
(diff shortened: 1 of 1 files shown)


app/misc/authenticate.ts
```diff
@@ -139,7 +139,7 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
-      else if (err == "Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443"){
+      else if (err.toString().indexOf("ENOTFOUND") != 0 || err.toString().indexOf("ENOENT") != 0){
```
colinwerner: It would be nice if we could somehow check err.errno or something similar? If not possible, perhaps we should make it a TODO. Let me know the results?

Performing multiple string comparisons is inefficient :(; however, it looks like this is what others have done in this file before.

app/misc/authenticate.ts
```diff
@@ -139,7 +139,7 @@ function getUserInfo(callback) {
...
         displayModal("No internet connection");
```
colinwerner: Can we update this dialog to be more specific? Perhaps "Unable to Authenticate: No internet Connection" -- assuming that this is the ONLY case that would see this dialog?

app/misc/authenticate.ts
mehrler12: err.errno I think will work! Can you tell me what it returns on mac? I think it'll still require the multiple string comparisons but we can at least get rid of the indexof operation.

app/misc/authenticate.ts
mehrler12: From what I've read on this error message it's specfically a DNS thing so how about "DNS not found: Check your internet connection"?

mehrler12: It seems like ENOENT is a bit more generic actually, it just means there is nothing there. Do we want a different error message for that one? "Authentication page not found: Check your internet connection"?

colinwerner: Yes, I think "Authentication Error: Please check your internet connection?".

app/misc/authenticate.ts
```diff
@@ -139,7 +139,7 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
-      else if (err == "Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443"){
+      else if (err.errno == "ENOTFOUND" || err.errno =="ENOENT"){
```
colinwerner: Random Question: Is there a precedent for constant strings?

mehrler12: The precedent in the project is this.

Unimax:
I was wondering we could have used navigator.onLine somehow (to confirm i guess) but I guess it's not needed for this issue. 
link: https://electronjs.org/docs/tutorial/online-offline-events

Issue #10, opened 2019-05-21
Pressing F5 resets the app


omazhary:
Regardless of where you are within the app, pressing F5 resets the whole app and takes you back to the login screen. This should not be possible as it may produce unexpected behavior for the rest of the app and gives the impression the app is not stable.

**Steps to reproduce:**

1. Sign in or continue without sign in on Visual Git
2. Press F5
3. Observe app reset

**Expected Output:**

F5 no longer takes you back to the login screen

colinwerner:
Is this actually a bug? Sounds like a built in feature to me, we should be able to override the F5 to be a no-op, but is that desired?

colinwerner:
We need to determine what the behaviour is for this.

Also, this would be easier if we do #19 first.

jerushachua:
This is assigned to group 2.

trstndvll:
@aeogunmuyiwa I tested your branch on my Mac as you asked, and I can confirm that fn+F5 which used to reset the app does not reset it anymore. That's awesome! I could still reset the app using command+R, so it looks like that still needs to be fixed. Hope this helps!

aeogunmuyiwa:
Thanks @trstndvll I appreciate your feedback.

Unimax:
whoever doing the testing please also test when u switch to another window (say using sift+tab and switch to chrome) and later coming back to our app it still should block all the shortcuts to refresh.

aeogunmuyiwa:
Please can someone pull the latest branch of bug-issue10. I have updated the code so it should not crash when you switch tabs and also please test the F5 and refresh short-keys. Please leave your feedback as well. Thanks.

trstndvll:
@aeogunmuyiwa  Just pulled it again and I can confirm command+R isn't working anymore. Great job guys!

Unimax:
I tested on Windows 10 & Ubuntu 18, Both Ctrl+R and F5 is no longer refreshing/Crashing the application.

aeogunmuyiwa:
Thanks @trstndvll @Unimax  for testing it out and confirming that it works properly now. I believe the issue is done , I will confirm what steps we need to take before we close the issue.

aeogunmuyiwa:
closing.

Unimax:
> closing.

just for future times, an issue is closed after merging the PR. read the wiki colin had made.
anyway good work on the issue and thanks for being there while Colin was reviewing.

aeogunmuyiwa:
[...]

PR #59, opened 2019-05-24
Fix Issue #10 : using globalShortcut to overriding the default function of 'F5' and 'Ctrl+R'

aeogunmuyiwa:
#10 The app crashes where caused by default electron shortcut-key settings. I have registered the refresh keys(F5 & CmdOrControl +R) so it no longer takes you back to the login screen. The bug also occurred when you switched pages and changed focused screens so I fixed that as well.


Unimax:
app/misc/authenticate.ts, app/misc/git.ts, and app/misc/graphSetup.ts contains minor changes not related to this issue.  can anyone confirm that I need to revert them and put them as a separate issue or this is fine? (because those are small changes)

jerushachua:
The changes are those small fixes with brackets here and there right? It's pretty small so you can just leave it as is.

colinwerner:
The small changes are not a big deal, but since we're supposed to be following agile principles I will provide some more information. If we are strictly following agile principles, then you should not sneak in other changes. The reason is that normally we would assign a value to an issue (say 5 points) and at the end of the sprint we would add up all the points we completed and that is our velocity. Since you spent time on these other changes (however trivial they are) we do not gain any velocity for this sprint from them, thus it is hard for us to track what our actual velocity could be. Make sense?

index.js
```diff
@@ -4,10 +4,14 @@ const electron = require('electron');
 const app = electron.app;
 const Menu = electron.Menu;
 const BrowserWindow = electron.BrowserWindow;
+const { globalShortcut } = require('electron')
```
colinwerner: I don't know TS very well. Can you explain what the difference is between line 6 and 7? Obviously they are syntactically different; however, do they accomplish the same thing?

I'm curious if lines 3-7 may be changed to:
const { app, Menu, BrowserWindow, globalShortcut } = require('electron')?

index.js
```diff
@@ -18,12 +22,29 @@ function onClosed() {
 	mainWindow = null;
 }
 
+function overrideRefreshWindowShotcuts(){
+	/*const ret = */globalShortcut.register('CmdOrCtrl+R', () => {
+		//console.log('CmdOrCtrl+R')
```
colinwerner: Is this basically saying on refresh do nothing? If so perhaps a comment would be beneficial?

index.js
```diff
@@ -18,12 +22,29 @@ function onClosed() {
...
+	console.log(globalShortcut.isRegistered('F5'))*/
+}
+
 function createMainWindow() {
 	const win = new electron.BrowserWindow({
 		backgroundColor : "#000",
 		icon: __dirname + "/assets/icons/Icon.png"
 	});
-
+	//electronLocalshortcut.unregisterAll(win);
```
colinwerner: I'd suggest removing this; otherwise we have no idea why it is commented out?

index.js
```diff
@@ -18,12 +22,29 @@ function onClosed() {
...
+function overrideRefreshWindowShotcuts(){
+	/*const ret = */globalShortcut.register('CmdOrCtrl+R', () => {
+		//console.log('CmdOrCtrl+R')
+	})
+	/*const ret1 = */globalShortcut.register('F5', () =>  {
+		//console.log('F5')
+	});
+
+	/*if (!ret || !ret1) {
+		console.log('registration failed')
```
colinwerner: I think throwing an error (or at least logging one) here would be beneficial as it is unexpected behaviour?

index.js
aeogunmuyiwa: Yes, It is basically  doing that . I will leave a comment on it saying what it does.

index.js
aeogunmuyiwa: I will change the function and log it if an expected behavior occurs, however it is guaranteed that it will register the shortcut once the app is open per operating system .

index.js
aeogunmuyiwa: They are equivalent ways of assignment, "const { globalShortcut } = require('electron')" is just shorter . I have updated the code so it uses const { app, Menu, BrowserWindow, globalShortcut } = require('electron'). Thanks

colinwerner: Ok, sounds good. Thank you for the clarification.

colinwerner:
Please ensure you capture a video of the behaviour before your change, please.

Issue #47, opened 2019-05-22
Unable to see edited files (staged or unstaged) in visualgit


colinwerner:
When I make a change to a local repo (whether staged or not), I am unable to see the edited files in the repo.

Steps to reproduce:
1.) Open local repo
2.) Edit a file
3.) Save, Close

Expected Steps:
4.) the edited file should be visible

Actual Steps:
4.) the edited files are not visible

![image](https://user-images.githubusercontent.com/15122789/58147210-475a5700-7c0e-11e9-82ef-db672c494a45.png)

colinwerner:
This is related to #48.

colinwerner:
Actually I can see the staged/unstaged files (well some of them) when there are more than one.

Also, this issue seems specific to Mac.

jerushachua:
This issue will be assigned to group 4.

colinwerner:
I can fix the initial issue. However, there appears to be another issue whereby files staged via command line are not recognized in visual git. *humpf*

Here's the diff for anyone who wants to look.
[diff.txt](https://github.com/uvicgse/project-virtual-team-1/files/3210468/diff.txt)

colinwerner:
My changes are in issue-#47-unable-to-see-edited-files -- we can look at the staged vs non-staged issue later.

PR #60, opened 2019-05-24
Fix Issue #47: do not use '//' as this won't work on *NIX, so use path.join instead

colinwerner:
Fixed Issue #47, which only occurred on non-Windows platforms.

The root cause of the issue is that a '/' was being used in the file path, which is not understood by *NIX.

I used the path.join library to make the path, which knows whether to use a '/' or a '\'.

### DIFF
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -825,7 +825,7 @@ function displayModifiedFiles() {
           let filePaths = document.getElementsByClassName('file-path');
           for (let i = 0; i < filePaths.length; i++) {
             if (filePaths[i].parentElement.className !== "file file-deleted") {
-              let filePath = repoFullPath + "\\" + filePaths[i].innerHTML;
+              let filePath = path.join(repoFullPath, filePaths[i].innerHTML);
               if (!fs.existsSync(filePath)) {
                 filePaths[i].parentElement.remove();
               }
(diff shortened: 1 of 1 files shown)


jerushachua:
Have you checked if this change affects Windows or Linux users in any way?

colinwerner:
It has been tested on both Windows and MacOS. 

@Unimax are you able to test this on Linux, please?

Unimax:
I tested on Ubuntu 18.04.2 LTS and its working. i can see single and multi-file changes as expected.

colinwerner:
@Unimax I only replaced the instances in this particular file. There are other instances that might need to be replaced.

@maxUser are you interested in looking at these?
project-virtual-team-1 $ grep "\\\\" -nrFo --include=*.ts .
./app/misc/wiki.ts:23:\\
./app/misc/wiki.ts:26:\\
./app/misc/wiki.ts:50:\\
./app/misc/wiki.ts:82:\\
./app/misc/wiki.ts:140:\\
./app/misc/repo.ts:207:\\
\\
./app/misc/git.ts:509:\\
./app/misc/git.ts:1209:\\
./app/components/addRepository/add.repository.component.ts:90:\\
./app/components/addRepository/add.repository.component.ts:98:\\

colinwerner:
PS: I'm still waiting for one more +1

Issue #51, opened 2019-05-22
Allow a user to view current tags in current repo


colinwerner:
This is a decomposition of #39.

view current tags in repo (git tag)

sterlinglaird:
We can break this issue into 2 distinct tasks:

1. View optional textual data next to each commit. This is the UI component of this issue and would involve adding some way of viewing data for each commit. Since the text will be an optional tag, no additional text-box will appear if the data isn't available. The design of how this would be displayed would be part of this task.
2. Receiving tags from git. This is the back-end portion for interfacing directly with git.

Once both tasks are complete the UI elements can be linked to the tags that are received from git.

colinwerner:
@sterlinglaird I believe this task is to show the results of running 'git tag' on the current branch. 

Showing the tags on each commit is #54.

sterlinglaird:
When VisualGit queries for git references (to get branches), it gets all the tags as well. The system doesn't recognize that there is a difference between the 2 reference types so it currently displays all tags as branches in the GUI. We will need to parse the tags from the branches and keep a separate list of them so the GUI doesn't treat tags as branches.

sterlinglaird:
@colinwerner I understand now, we didn't notice that was a follow-up task. We will add a drop-down list (like the branch list) for the tags present the repo.

colinwerner:
@sterlinglaird interesting. I've never used tags in git before; however, it sounds like there is a.) a small bug to do and b.) a new feature to implement.

As you suggested, after performing the query we can separate them into branches and tags and add a new dropdown to show the tags?

Good investigation!

sterlinglaird:
We have updated the branch dropdown to now contain tabs that can be used to switch between a list of branches and a list of tags in the repo. The search/create branch textbox doesn't do anything at the moment and was inactive before this change. We imagine this is a bug that will be addressed at some point. Currently clicking on a tag name does nothing, but in a following issue (#61) we will add support for checking out a tag just like a branch.

Here are the new branch and tag dropdown menus:
![branch](https://user-images.githubusercontent.com/13267819/58374135-7f52ea00-7eed-11e9-94b4-00dee21f2a9e.PNG) 
![tag](https://user-images.githubusercontent.com/13267819/58374136-837f0780-7eed-11e9-9c04-520ace4243b5.PNG)

Issue #39, opened 2019-05-21
Add tagging support


omazhary:
**Feature request:**

Add support for tagging commits in the past and/or the current commit that has yet to be pushed. This will allow users to mark important points in their repository for example "v1.0" without needing to search for a certain commit message/checksum. Users should also be able to view the tags on the main page

colinwerner:
The tagging feature should probably be broken down into separate chunks (https://git-scm.com/book/en/v2/Git-Basics-Tagging):
1.) view current tags in repo (git log)
2.) search current tags in repo (git log -l "v1.0")
2.) create a new tag (two different types) (git tag -a v1.4 -m "my version 1.4" and git tag v1.4-lw)
3.) showing existing tags associated with commits
4.) add a tag to a commit
5.) delete a tag from the repo

jerushachua:
This issue will be decomposed into the above issues and we will push out all issues except one.

jerushachua:
The first issue will be assigned to group 1.

hdauvic:
This feature addition is scheduled to be completed in Sprint 4.

jerushachua:
Closing as the remaining issues related to tagging have their own documentation.

PR #62, opened 2019-05-25
#ISSUE 51: allow user to view tags - New Feature

SiRumCz:
New tagging feature:
- redesigned the Branch dropdown menu UI to allow user to also view the lightweight and annotated tags.
- now branch list and tag list are seperated into two tabs.
- added placeholder for tag search (maybe and create) function.


app/misc/repo.ts
```diff
@@ -2,7 +2,7 @@ let Git = require("nodegit");
 let repoFullPath;
 let repoLocalPath;
 let bname = {};
-let branchCommit = [];
```
sterlinglaird: branchCommit was completely unused and was within the code we worked on, so I took the liberty to remove it

app/misc/repo.ts
```diff
@@ -261,7 +261,9 @@ function openRepository() {
   function refreshAll(repository) {
     document.getElementById('spinner').style.display = 'block';
     let branch;
-    bname = [];
+    bname = {};
```
sterlinglaird: I dont think this affected the code at all, but bname should have been a object, not an array.

app/misc/repo.ts
```diff
@@ -274,41 +276,55 @@ function openRepository() {
         //Get the list of branches from the repo
         return repository.getReferences(Git.Reference.TYPE.LISTALL);
       })
-      .then(function (branchList) {
```
sterlinglaird: Many of the changes in this section are just changing names from `branch.*` to `ref.*` to reflect the fact that these variables are in fact git refs and not specifically branches

SiRumCz:
As we were testing this feature, we found out VisualGit won't be able to load any git repositrory that contains annotated tags. This is a bug before the implementation of this feature.

app/misc/git.ts
```diff
@@ -282,7 +282,9 @@ function getAllCommits(callback) {
         function (cb) {
           if (!refs[count].isRemote()) {
             console.log("referenced branch exists on remote repository");
-            repos.getReferenceCommit(refs[count])
+            refs[count].peel(Git.Object.TYPE.COMMIT)
```
sterlinglaird: Annotated tabs used to cause failures here because they are not simply a reference, like non-annotated tags are. The peel(Git.Object.TYPE.COMMIT) function isolates the commit reference, exactly what we want.

SiRumCz: Tthis is a good fix, I can open a testing repo with annotated tags now.

colinwerner:
We need to figure out a way to refresh the list of tags. Here is what I did:
1.) opened a repo (with no tags)
2.) looked at the tag tab (there was nothing -- perhaps we should say "No tags available"?)
3.) created two tags from the command line
4.) went back to the tab and still saw nothing

After reloading the repo, I could see the tags.

colinwerner:
Also, when clicking a tag, it does nothing. Should we create a feature request to do something?

SiRumCz:
> We need to figure out a way to refresh the list of tags. Here is what I did:
> 1.) opened a repo (with no tags)
> 2.) looked at the tag tab (there was nothing -- perhaps we should say "No tags available"?)
> 3.) created two tags from the command line
> 4.) went back to the tab and still saw nothing
> 
> After reloading the repo, I could see the tags.

Live reload should be a new feature for the future developement as this feature also applys to the branch list.

SiRumCz:
> Also, when clicking a tag, it does nothing. Should we create a feature request to do something?

This pull request is a implementation to issue #51 which is a decomposition of issue #39 . click interaction will be a feature in future version.

colinwerner:
Thanks @SiRumCz -- can you please ensure that these are accurately captured in an issue and reference the numbers here for future/past traceability, please?

SiRumCz:
@colinwerner Absolutely. Thanks for your suggestion.

Issue #26, opened 2019-05-21
Issue with selecting folder when cloning -Path Does not Exist


omazhary:
**Steps to reproduce**

1. Click on Save without any URL
![image](https://user-images.githubusercontent.com/22312351/57658580-56992f00-7594-11e9-9254-461f243f77a7.png)

2. It will give a pop up message saying "Clone Failed" close it
![image](https://user-images.githubusercontent.com/22312351/57658609-703a7680-7594-11e9-9e9a-9f40b8da8aa7.png)

3. Go back to the main screen and click on clone again in the same folder. (You get this error)

![image](https://user-images.githubusercontent.com/22312351/57658624-76c8ee00-7594-11e9-9638-58293a24497a.png)

But if you choose a different folder it does not give you that error

Athenasko:
Steps for Reproduction (Differ slightly from above as it seems new functionality has been added since then):

My steps:
1. Click Browse button beside "File location to save to".
2. Select a folder to save to.
3. Close pop-up telling you to enter repository URL to clone.
4. Click Browse button again.
5. Click "Select folder" without changing anything.

Expected Results:
1. If no repository URL is entered, prompt should appear again to enter URL to clone from.
2. If repository URL is entered, file location should appear.

Actual Results:
"Path Does Not Exist. Check path again." error alert appears.

jerushachua:
This issue is assigned to group 4.

jpstoldt:
After further analysis, we have identified the underlying issue as a bug in Electron which affects all directory selection dialogs using the INPUT TYPE=FILE html tag in the application. Options to resolve are 1) change the user interface flow of the application to avoid the issue, however this would result in a loss of current application functionality; 2) replace uses of INPUT TYPE=FILE tag with a native system dialog function in Electron which does not have the issue; 3) upgrade Electron to its latest version which contains a bug fix to the issue.

Issue #23, opened 2019-05-21
Update to newest NodeJS


omazhary:
Blocks issue #98.

Update from 6.11.0 to 10.15.3. Brings a lot of improvements and security patches to NodeJs. Needed to update angular

jpstoldt:
Resolved with PR #63

PR #63, opened 2019-05-27
Fix Issue #26: Upgrade Electron and Node versions

jpstoldt:
Upgrade of Electron to version 5 and the latest Node version resolves this file system dialog issue. Installation of Python 2 is required for rebuilding native node modules as part of this upgrade (https://github.com/uvicgse/project-virtual-team-1/wiki/Upgrade-Process)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,4 @@
 runtime = electron
-target = 1.2.8
+target = 5.0.0
 target_arch = x64
 dist_url = https://atom.io/download/atom-shell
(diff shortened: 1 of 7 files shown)


colinwerner:
This also fixes #23?

jerushachua:
Is installation of python 2 also a dependency for the first installation with a clean repo?

jpstoldt:
> This also fixes #23?

It does - will change the reference to show that we are fixing #23 and #26

jpstoldt:
> Is installation of python 2 also a dependency for the first installation with a clean repo?

yes, you need Gyp which is part of Python 2 to recompile native node modules for Electron with a clean repo as well  (https://electronjs.org/docs/tutorial/using-native-node-modules)

PR #64, opened 2019-05-27
removed inappropriate log message for bug-issue10

aeogunmuyiwa:
I removed the  inappropriate log message for bug-issue10, The "registration failed " messaged was removed because it served no purpose and was initially added for testing reasons.


trstndvll:
On Mac, I can no longer use command+R or fn+F5 to reset the app to the login screen. I also cannot see the console log that was taken out, so things look good to me! Great job.

aeogunmuyiwa:
Thanks @trstndvll  @sterlinglaird . I will merge the changes, move the issue to done and close the issue.

PR #67, opened 2019-05-28
Bug 30 - Pushing changes before committing crashes application

trstndvll:
Current fix: If you attempt to push without committing, a dialog block will explain that you cannot push as there are no commits, and will not allow the push function to continue. This is true if you have either un-staged changes or staged changes. If you have committed changes, you will be allowed to push. 

The behaviour of the app crashing after the second push still occurs in our repo as it has not been re-based with the other group's changes where this issue has been fixed (the app will crash after 2 times even with pushes that have commits).

Future plans: Currently, if you commit changes in VisualGit, then exit the app and re-open, the commits will not show and you cannot push because of our check. We would like to solve this case in sprint 2 by checking the diff between local and remote repos, instead of relying on a check based on actions in the app. This will involve more research into nodegit functions.

### DIFF
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -102,7 +102,7 @@ function stage() {
         }
       }
       if (filesToStage.length > 0) {
-        pushButNoCommit = 1;
+        pushButNoCommit = 0;
         console.log("staging files");
         stagedFiles = index.addAll(filesToStage);
       } else {
@@ -117,6 +117,7 @@ function stage() {
       filePanelMessage.parentNode.removeChild(filePanelMessage);
     }
   }
+  pushButNoCommit = 0;
 }
 
 function addAndCommit() {
@@ -221,6 +222,7 @@ function addAndCommit() {
         updateModalText("You have not logged in. Please login to commit a change");
       }
     });
+    pushButNoCommit = 1;
 }
 
 function clearStagedFilesList() {
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -14,7 +14,7 @@ let theirCommit = null;
 let modifiedFiles;
 let warnbool;
 var CommitButNoPush = 0;
-let PushButNoCommit = 0;
+let pushButNoCommit = 0;
 let stagedFiles: any;
 let vis = require("vis");
 let commitHistory = [];
@@ -102,7 +102,7 @@ function stage() {
         }
       }
       if (filesToStage.length > 0) {
-        PushButNoCommit = 1;
+        pushButNoCommit = 1;
         console.log("staging files");
         stagedFiles = index.addAll(filesToStage);
       } else {
@@ -391,7 +391,7 @@ function pushToRemote() {
   let branch = document.getElementById("branch-name").innerText;
 
   // Check to make sure we are not pushing without first committing files
-  if (PushButNoCommit == 0) {
+  if (pushButNoCommit == 0) {
     console.log("There are no staged files")
     displayModal("You have not committed, your push cannot be empty.");
     return;
@@ -419,6 +419,7 @@ function pushToRemote() {
             })
             .then(function () {
               CommitButNoPush = 0;
+              pushButNoCommit = 0;
               window.onbeforeunload = Confirmed;
               console.log("Push successful");
               updateModalText("Push successful");
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -394,8 +394,8 @@ function pushToRemote() {
 
   // Check to make sure we are not pushing without first committing files
   if (pushButNoCommit == 0) {
-    console.log("There are no staged files")
-    displayModal("You have not committed, your push cannot be empty.");
+    console.log("There are no commits available to push")
+    displayModal("You have not committed any files, your push cannot be empty.");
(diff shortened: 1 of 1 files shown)


app/misc/git.ts
```diff
@@ -387,6 +391,14 @@ function pullFromRemote() {
 
 function pushToRemote() {
   let branch = document.getElementById("branch-name").innerText;
+
+  // Check to make sure we are not pushing without first committing files
+  if (pushButNoCommit == 0) {
+    console.log("There are no staged files")
```
colinwerner: This error message is a bit misleading. Shouldn't it be "There are no commits available to push"?

trstndvll: Good point @colinwerner ! I've updated the message to reflect your suggestion. My group will make sure to create a new issue after class.

sterlinglaird:
I'm a little unsure about this fix because you can push other things besides commits, such as tags or branches. We need to make sure we can still push those without needing to commit files too.

trstndvll:
> I'm a little unsure about this fix because you can push other things besides commits, such as tags or branches. We need to make sure we can still push those without needing to commit files too. Really, Git allows you to push without having _anything_ changed so we should be able to find the actual source of the crash and allow users to push whenever.

@sterlinglaird Thanks for looking into this. Since tags are not implemented yet (correct?), would that be something we need to add functionality for in a future issue? And in terms of the branches, could you just explain what you mean by pushing branches? Do you mean fetching upstream, or like switching? Thanks!

sterlinglaird:
@trstndvll When you create a branch in git, it only exists locally on your machine until you do a `git push`. The same goes for tags, amending commit messages, and probably others that I can't think of. In general I think we should be able to push at anytime because until you push you can't ensure the state of the upstream repo matches what you have locally.

trstndvll:
That totally makes sense, @sterlinglaird . I have just added a check that sees if you have any branches not existing on remote so you can push, so I can commit that to this PR. Since we're going to make another issue to work on cases that this didn't cover, would you think it would be okay to add tag functionality to that too? I'm just thinking we should check in to see exactly what functionality he wants, since there is what git should do, but also since this is meant to teach people how to use git.

In any case, I'm going to meet with my group and see what we think in terms of demo tomorrow.

sterlinglaird:
I'm not sure checking if there are any new branches is sufficient because even existing branches can move without a commit (or branches can be deleted, moved through a commit that was never pushed etc). And besides, there's still the issue with amended commit messages and tags. We need to keep in mind that arbitrary commands can be made using the git command line, and VisualGit should be able to handle those. Because the changes in this PR remove functionality (arbitrarily pushing upstream) I'd be hesitant to merge it. Is there any way to allow VisualGit to operate the same way as before (allowing you to push always) but just fix the crash itself instead of restricting the user interaction?

trstndvll:
@sterlinglaird that's a good point. @colinwerner just to update you, my group and I just had a meeting and we discussed how this issue has progressed. The original scope of this ticket was that pushing changes that do not have any commits was crashing the application. Although we have fixed that issue with our work, it does affect functionality as @sterlinglaird says. 

We just want to make sure that you and James know that we've been doing a lot of work on this issue this week (testing, research, implementation, edge case discovery, etc.), and it seems like this pushing crash has gotten out of the scope of this original issue. At this point we think the push function as a whole may need to be re-worked to solve this crashing issue, and not just the empty commit issue. 

For the demo tomorrow, we do not want to disrupt any of the other groups with our tests on our branch, as we think there may need to be a couple new issues created from this original one. Should we hold off on merging? @colinwerner would you be able to weigh in on what you think the best way forward is for tomorrow? I think we need to speak with James to pinpoint what exact functionality he expects.

@colinwerner Feel free to reach out to me on Slack if you have any questions. Thanks for your help.

SiRumCz:
I totally agree with what @sterlinglaird was saying regarding the arbitrary behaviors of git push. I think you guys should instead of checking there is any unpushed local commits, maybe try to see if you can use something similar like a try catch block to try to call push function and catch any ERROR message nodegit responds with and do the operations from there.

colinwerner:
> For the demo tomorrow, we do not want to disrupt any of the other groups with our tests on our branch, as we think there may need to be a couple new issues created from this original one. Should we hold off on merging? @colinwerner would you be able to weigh in on what you think the best way forward is for tomorrow? I think we need to speak with James to pinpoint what exact functionality he expects.
> 
> @colinwerner Feel free to reach out to me on Slack if you have any questions. Thanks for your help.

Given the time left, I say we hold off. In the meantime, please prepare to justify delaying this feature (I think you have just described everything that needs to be said). 

Also, do you know why the app crashed? That really is what we are trying to solve, isn't it?

trstndvll:
@colinwerner Thanks for your quick reply, I think that would be best to hold off. In terms of why the app is crashing, we are still looking into why it occurs. It seems to still crash even after two pushes that have commits, so it must be crashing in a more broad sense. We are continuing to look into the push function itself. I know my group-mate @aeogunmuyiwa had attempted this weekend to re-work the push function, so I'm not sure if he could speak more to that.

trstndvll:
moving to sprint 2

PR #75, opened 2019-05-30
Issue 14 invalid creds shows unclear error

colinwerner:
check the err object for a statuscode and if it is 401 (unauthorized) show an appropriate dialog to the users.

I also made all the {} in this block consistent with each other.

### DIFF
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -142,10 +142,10 @@ function getUserInfo(callback) {
       else if (401 == err.statusCode) {
         displayModal("Authentication Error: Please check your username and password.")
       }
-      else if (err.errno == "ENOTFOUND" || err.errno =="ENOENT"){
+      else if (err.errno == "ENOTFOUND" || err.errno =="ENOENT") {
         displayModal("Authentication Error: Please check your internet connection");
       }
-      else{
+      else{ 
         displayModal(err);
       }
       document.getElementById('grey-out').style.display = 'none';
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -139,9 +139,13 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
+      else if (401 == err.statusCode) {
+        displayModal("Authentication Error: Please check your username and password.")
+      }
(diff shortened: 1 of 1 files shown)


colinwerner:
I realize this change might be negated by the oauth work.

roweba:
I feel like this error message isn't very user-friendly, I feel like a better error message would be "Incorrect Username or Password. Please try again." All the logic for catching the error looks good.

app/misc/authenticate.ts
```diff
@@ -139,9 +138,13 @@ function getUserInfo(callback) {
           $("#otpModal").modal('show');
         });
       }
-      else if (err.errno == "ENOTFOUND" || err.errno =="ENOENT"){
+      else if (401 == err.statusCode) {
```
jerushachua: Is there a reasoning as to why err.statusCode is used here instead of err.errno, which is the error checking used for the Auth case below?

colinwerner: Unfortunately, it's an entirely different object (HttpError), so the actual error is stored in a different variable name, so this is all we can do!

colinwerner:
> I feel like this error message isn't very user-friendly, I feel like a better error message would be "Incorrect Username or Password. Please try again." All the logic for catching the error looks good.

I was just following suite of what else was there, which seemed to pass muster. Also, note that we can assume anyone using VisutalGit (and git) should be fairly technical in nature?

I'll ask James.

roweba:
> > I feel like this error message isn't very user-friendly, I feel like a better error message would be "Incorrect Username or Password. Please try again." All the logic for catching the error looks good.
> 
> I was just following suite of what else was there, which seemed to pass muster. Also, note that we can assume anyone using VisutalGit (and git) should be fairly technical in nature?
> 
> I'll ask James.

That's fair! I just think that a UI should be designed to be as simple as possible for the user. And I feel like you could argue that someone who uses VisualGit may not be that technical, hence the need to use a Git GUI. Either way it's not too big of a deal

app/misc/authenticate.ts
jerushachua: Oh I see - makes sense.

colinwerner:
Confirmed with James that this message is fine.

Issue #74, opened 2019-05-30
File editor button text overflows


colinwerner:
The file editor button text overflows the button outline.
I don't believe this occurs on all setups.
I'm on windows @ 2736 x 1824 resolution.
Also occurs when I change to 1920 x 1080 resolution.

image

Steps to reproduce:

Start application
clone or open a repo
Observe text overflow in the bottom left of the screen

colinwerner:
This is issue 51 in project-seed.

PR #76, opened 2019-05-30
CSS fix for the editor button (issue #74)

subahm:
The "file editor" text stays inside the button and those three buttons at the footer doesn't disappear in the left side for a smaller window size.
Stopped the white space from getting wrapped and fixed the position of the three footer buttons so they don't disappear on the left side for smaller windows

### DIFF
--- a/stylesheets/styles.css
+++ b/stylesheets/styles.css
@@ -217,8 +217,8 @@ body .navbar .dropdown-menu .select-menu-list .list-group-item:hover {
 }
 
 #visual-git-logo{
-  width: 150px; 
-  height: 150px; 
+  width: 150px;
+  height: 150px;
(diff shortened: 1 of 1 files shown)


subahm:
Do we want to fix the wrapping for "Stage All" and "Unstage All" buttons as well with this?
Screen Shot 2019-05-29 at 9 49 54 PM

colinwerner:
I'm not the best reviewer for CSS issues, although your changes seem fine to me.

@subahm: if it's not much effort to fix them, sure, why not? However, I'd like to hear from someone who knows CSS better than me ;-P

Fixed Issue #74 for easy reference.

Unimax:
this should also close issue #2

PR #77, opened 2019-05-30
Issue 8 - Navbar Icons Adjusting Irregularly

roweba:
The issue here seemed to be with the bootstrap grid system having a four-column width, which messed with the icons upon resize. Issue #8

### DIFF
--- a/app/components/header/header.component.html
+++ b/app/components/header/header.component.html
@@ -59,7 +59,7 @@
       </ul>
 
       <!--Toolbar-->
-      <ul class="navbar-nav col-md-6 hidden-xs toolbar-Container">
+      <ul class="navbar-nav col hidden-xs toolbar-Container">
         <li class="toolbar-Icons" onclick="openWiki()">
           <i id="wiki-icon" class="fa fa-wikipedia-w fa-2x"></i>
           <span class="toolbar-Text">Wiki</span>
--- a/app/components/header/header.component.html
+++ b/app/components/header/header.component.html
@@ -59,7 +59,7 @@
       </ul>
 
       <!--Toolbar-->
-      <ul class="navbar-nav col-md-4 hidden-xs toolbar-Container">
+      <ul class="navbar-nav col-md-6 hidden-xs toolbar-Container">
         <li class="toolbar-Icons" onclick="openWiki()">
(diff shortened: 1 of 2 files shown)


colinwerner:
Seems simple, but I'll let others who are more knowledgable in this area to comment. (I will help run it through a gauntlet of tests though :)).

roweba:
> this won't solve the extra padding getting added while hovering. right?

It doesn't, it simply solves the problem of the icons adjusting themselves to two rows. I can look into fixing this problem too.

roweba:
> I recommend looking into fixing the other issue regarding button resize first and putting all your changes into one pull request. It makes for easier reviewing and cleaner merging.

Thanks for the input. When I fix the other issue and commit it to this branch won't it already be in the same pull request?

jerushachua:
Sure, you can just update this pull request with your changes when they're ready :)

roweba:
Changes have been added!

jerushachua:
Thanks! I'll test this tomorrow and approve if everything looks good.

roweba:
> i would also like a more color change when we hover the icon. to make it more clear which one is currently selected. what other think?
> also, there is still a 10px difference when we hover right? is that behavior needed?

Changing the color on hover could be easily implemented but I think that's beyond the scope of this issue. The reason for that 10px difference on hover is because the hover effect adds extra padding around the icon so there needs to be more space. I think that also might be beyond what this issue is trying to address but I can ask James.

hdauvic:
Everything looks good for this specific issue, but it does bring up a possible nit-picky issue that at certain window sizes (as I make the window smaller and smaller) the sign in button will drop below the menu buttons, and then back up, and then back down. As well, the windows sizes at which it does this seem to be quite random. Is that something worth addressing?

roweba:
It does that because once you get to a certain resolution there isn't enough space for the icons and the sign in button to be in-line. This pull request didn't make that happen, if you look on master it's doing the same thing.

roweba:
@hdauvic I think I misunderstood when I first read this, I see what you were pointing out. I can look into this

roweba:
@hdauvic I fixed the problem you were addressing, can you do one last check before I merge this?

hdauvic:
Check worked great! Thanks for fixing this.

PR #78, opened 2019-05-30
fixed problem with get current branch name after added label 'Branch: '

SiRumCz:
This should fix the problem caused by concating 'Branch: ' label with real branch name.

### DIFF
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -326,7 +326,7 @@ function PullBuffer() {
 
 function pullFromRemote() {
   let repository;
-  let branch = document.getElementById("branch-name").innerText;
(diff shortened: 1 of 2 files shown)


SiRumCz:
@Unimax very interesting idea, but in my understanding of tag, it is more like a pointer to a certain commit ranther than a link to a branch.

Unimax:
> @Unimax very interesting idea, but in my understanding of tag, it is more like a pointer to a certain commit ranther than a link to a branch.

i think you misunderstood my commend so i updated it. you are right :)

SiRumCz:
@Unimax yea, I see your point. Probably need a more precise naming scheme for the future implementation. 

@Unimax @trstndvll thanks!

Issue #66, opened 2019-05-28
Need a live reload feature for both branch list and tag list


SiRumCz:
As user creates tag or branch using command line interface(outside of VisualGit), VisualGit should recognizes the change and also live-reloads the list. This issue is a follow up of issue #51 which is a decomposition of issue #39 .

SiRumCz:
pull request #79

SiRumCz:
Fixed in Pull Request #79 .

Issue #54, opened 2019-05-22
Allow a user to see the tags on existing commits


colinwerner:
This is a decomposition of #39.

showing existing tags associated with commits

hdauvic:
Since this is associated with Issue #51 I started this issue once clarification was given. So far I've added in code framework to support the graphical GUI showing tags associated with each commit.
TODO:
- Create get all tags function, which is likely associated with Issue #52 
- figure out best console.log message to display tags along side commit id and message
- see how tags display in comparison to branches; branches have a short name utilized in the graph, will tags need the same?

sterlinglaird:
Implemented in #84

PR #79, opened 2019-05-31
Issue #66: live-reload function for dropdown list - New feature

SiRumCz:
- added a new live-reload feature for keeping dropdown list up-to-date.
- fixed style issue of the dropdown list outline.
- clearBranchElement() now becomes clearBranchAndTagElement()

my implementation to the live-reload:
- add a monitor to track the list of reference(branch and tag), once it detects change to the list, it will refresh the list.
- the monitor tracks list every one second.

### DIFF
--- a/app/misc/repo.ts
+++ b/app/misc/repo.ts
@@ -260,7 +260,7 @@ function openRepository() {
   }
 
 // works as a monitor for any change to the reference list
-function refreshList() {
+function refreshList(verbose) {
   Git.Repository.open(repoFullPath)
     .then(function (repo) {
       repo.getCurrentBranch()
@@ -284,6 +284,7 @@ function refreshList() {
           tags = {};
           clearBranchAndTagElement();
           for (let i = 0; i < refList.length; i++) {
+            if (verbose) { console.log("reference name: " + refList[i].name()); }
             //get simplified name
             let refName = refList[i].name().split("/")[refList[i].name().split("/").length - 1];
 
@@ -293,6 +294,7 @@ function refreshList() {
                 // for remote branches add oid and branch name to remote branches map
                 remoteName[refName] = oid;
               } else if (refList[i].isBranch()){
+                if (verbose) { console.log(refName + ": adding branch to end of " + oid.tostrS()); }
                 // add to list of branches
                 if (oid.tostrS() in bname) {
                   bname[oid.tostrS()].push(refList[i]);
@@ -300,6 +302,7 @@ function refreshList() {
                   bname[oid.tostrS()] = [refList[i]];
                 }
               } else if (refList[i].isTag()){
+                if (verbose) { console.log(refName + ": adding tag to end of " + oid.tostrS()); }
                 // add to list of tags
                 if (oid.tostrS() in tags) {
                   tags[oid.tostrS()].push(refList[i]);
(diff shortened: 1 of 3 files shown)


app/misc/repo.ts
```diff
@@ -13,6 +13,7 @@ let span;
 let contributors: [any] = [0];
 let previousOpen;
 let repoName : string = "";
+let lastRefList = [];
```
SiRumCz: This variable serves the purpose of comparing newest fetched list to the last fetched one in refreshList().

app/misc/repo.ts
```diff
@@ -277,8 +352,8 @@ function openRepository() {
         return repository.getReferences(Git.Reference.TYPE.LISTALL);
       })
       .then(function (refList) {
-        let count = 0;
```
SiRumCz: Since this variable was not being used anywhere, I decide to delete it.

app/misc/repo.ts
```diff
@@ -277,8 +352,8 @@ function openRepository() {
...
-        clearBranchElement();
+        clearBranchAndTagElement();
+        refList.sort();
```
SiRumCz: This sort() was added to here to give a consistent order to the list.

app/misc/repo.ts
```diff
@@ -258,11 +259,85 @@ function openRepository() {
     }
   }
 
+// works as a monitor for any change to the reference list
+function refreshList() {
```
sterlinglaird: Is it possible to call refreshList() from refreshAll() so we don't have duplicated code for refreshing the list inside refreshAll()?

app/misc/repo.ts
```diff
@@ -258,11 +259,85 @@ function openRepository() {
...
+        .then(function () {
+          // grab the list
+          return repo.getReferences(Git.Reference.TYPE.LISTALL);
+        })
+        .then(function (refList) {
+          // sort refList alphabetically to get uniform order of the list
+          refList.sort();
+
+          // monitor any changes to reference list
+          if (lastRefList.length === refList.length && lastRefList.every(function(value, index) { return value.name() === refList[index].name()})) {
```
SiRumCz: This is a naive comparison of the two list of references.
Possibly, there could be a better solution than mine, but it passed all my test cases and gets the job done.

app/misc/repo.ts
SiRumCz: @sterlinglaird yes, I think it is a good idea and will look into it.

app/misc/repo.ts
```diff
@@ -273,61 +351,7 @@ function openRepository() {
         branch = branchParts[branchParts.length - 1];
       })
       .then(function () {
-        //Get the list of branches from the repo
```
SiRumCz: @sterlinglaird I redesigned the code based on your idea. Please have a look. Thank you.

SiRumCz: now that I think about it, I feel like I could simply just delete this part since there is already a dedicated function that checks the list.

sterlinglaird:
Looks good

index.html
```diff
@@ -98,8 +98,14 @@
   setInterval(function() {
     displayModifiedFiles();
   }, 3000);
+
+  // monitor changes on branch list and tag list, 1 check per second
+  setInterval(function () {
+    refreshList(false);
+  }, 1000);
```
colinwerner: Do we really need to monitor the branch and tag list every second? This seems wasteful.

Of course, I could be naive, but why don't you refresh the list when you open the dropdown?

Athenasko:
The live-reload works, however there are a couple of concerning behaviors:

- When switching between repos, the second switch always throws an error where it cannot add a new item. This does not seem to exist in other branches from what I've seen, which suggests this problem is introduced here. SwitchingBetweenReposCausesDuplicationIssue
- When using the branch tags to switch between branches, the bubbles that denote the branch names are erased. This also seems to be introduced here as the previous branch switching still kept the messages.

SiRumCz:
@Athenasko Thanks very much for your cncerns. Yes there are problems with the current drawing methods.

 As I am also collaborating with one of my group member on issue #54 , I believe these two problems were not introduced by this PR, and under certain circumstances they do exist on very first version of VisualGit. 

For the problem with first white dot node you mentioned as erased node, we are still investigating the reason in issue #54, hope they should be fixed with an other PR regarding that issue. So far I think this is some kind of duplicate nodes created in the first level of graph, but not sure why since it is not a 100% reproducible.

For the second one you mentioned that it tells you an id already exists. This is because in the original version there is an array named abstractList used to hold author information for generating second level graph, and when you open a new(second) repo, they didn't clean this array before the drawing so it caused that problem. I think the reason why everyone else didn't seem to encounter this problem is because they are likely working on test repos that only have one author and luckily dodged the bullet. However this problem is fixed in issue #54  and will be seen in the PR.

index.html
SiRumCz: @colinwerner yea, it does seem a little bit overkill and it also might affect performance. I will talk with my group members to discuss your idea. Thank you!

Athenasko:
Thanks for the clarifications @SiRumCz. Sounds like everything is worked out then. 

From what you said on my second point, it sounds like if both repos had the same author, is should have worked without issue for this ticket and only repos that had multiple authors should have cause this issue? Either way, glad it will be fixed.

SiRumCz:
@Athenasko I would assume so, but that array not only holds information about authors, but also under each author there is a list of commit ids these authors created. Therefore, it could potentially encounter any problem.

index.html
SiRumCz: Hi @colinwerner , 

After talked with my group members and heard other members' opinions. 
Below is my thoughts on the idea to this monitor:
The only drawbacks of this one check per second monitor that I could think of is that it could be slowing down the performance and using a little bit of resources. Since it is only a monitor, it will not refresh the list every one second, it just checks the list if no change, it will simply return. Base on this behavior, I would like to argue the performance will not be affected too much as I had also tested this on small and bigger project repos, and I didn't see any signs of performance change. 
However, The advantage of this monitor is that when user had the list open and he or she did some changes using CLI or other tools on the local repo, my monitor will detect changes immediately and show users the latest list of branches and tags. I think the pro of this monitor out weights the con.
But that is just my opinion on this part. Please feel free to correct me and share your opinions. 

Thank you.

colinwerner: Thanks @SiRumCz, that's fine.

PR #82, opened 2019-06-01
Issue 71 & 72 : double pull or push crash fix and  error handling

Unimax:
Solves Issue 71: pulling or pushing the 2nd time crashes the application. It looks like GitHub no longer let you use the same credentials object for different request requiring authentication. And hence we are creating and using a new object every time.

Solves Issue 72: in case of error occur during pull or push it shows that error message to the user. there was no catch in the process. (there was on in pull but at the wrong place) .

Solves new issue: errors during a pull cause the application to crash. for example, pull when there is no internet will crash the application with low-level assertion fail error. this issue is considered part of issue 72.

Note: no custom messages are created. the messages from the server are directly shown to the users.

### DIFF
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -80,8 +80,10 @@ function ModalSignIn(callback) {
         getUserInfo(callback);
 }
 
+// provide a fresh cred object. 
+// previously we were trying to use same object again and again which was cauing issues
 function getCredentials(){
-  // might change with Oauth
+  //Note: might change with Oauth
   return Git.Cred.userpassPlaintextNew(getUsernameTemp(), getPasswordTemp());
 }
 
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -160,7 +160,7 @@ function getUserInfo(callback) {
     encryptTemp(document.getElementById("username").value, document.getElementById("password").value);
   }
 
-  cred = Git.Cred.userpassPlaintextNew(getUsernameTemp(), getPasswordTemp());
+  cred = getCredentials();
 
   // Remove, once Oauth implemented
   client = github.client({
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -83,7 +83,7 @@ function ModalSignIn(callback) {
 // provide a fresh cred object. 
 // previously we were trying to use same object again and again which was cauing issues
 function getCredentials(){
-  //Note: might change with Oauth
+  //Note: might need to update with Oauth
   return Git.Cred.userpassPlaintextNew(getUsernameTemp(), getPasswordTemp());
 }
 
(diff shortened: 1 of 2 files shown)


app/misc/authenticate.ts
```diff
@@ -80,6 +80,12 @@ function ModalSignIn(callback) {
         getUserInfo(callback);
 }
 
+// provide a fresh cred object. 
+// previously we were trying to use same object again and again which was cauing issues
```
colinwerner: do we know why it was causing issues?

Unimax: found nothing from any changelog, but did found similar issues (4-5) many of them were unsolved (i did post my solution there too). one of them had the solution similar to what I came up with too. 
It looks like push/pull code might be calling some cred cleanup, and reusing the cred object is not possible. btw in the terminal also we have to enter login details every time right?

Unimax: how to find out why? I mean it was low-level segmentation fault error (assertion fail). similar error u can produce by unplugging ur internet and press pull. (also fixed) 
I mean I don't know how to go below to c level debugging in this application.

colinwerner: > found nothing from any changelog, but did found similar issues (4-5) many of them were unsolved (i did post my solution there too). one of them had the solution similar to what I came up with too.
> It looks like push/pull code might be calling some cred cleanup, and reusing the cred object is not possible. btw in the terminal also we have to enter login details every time right?

It depends on your env setup. That's fine, I was just wondering if we knew the underlying issue. And you do!

I would suggest adding your comments here in the code -- perhaps with a reference to an open issue with nodegit?

app/misc/authenticate.ts
```diff
@@ -80,6 +80,12 @@ function ModalSignIn(callback) {
...
+function getCredentials(){
+  //Note: might need to update with Oauth
+  return Git.Cred.userpassPlaintextNew(getUsernameTemp(), getPasswordTemp());
```
colinwerner: I would suggest syncing up with @dragonprevost so that we don't come across any hiccups once we merge.

app/misc/git.ts
```diff
@@ -395,27 +395,29 @@ function pushToRemote() {
...
-            });
+        }).then(function() {
+          CommitButNoPush = 0;
+          window.onbeforeunload = Confirmed;
+          console.log("Push successful");
+          updateModalText("Push successful");
+          refreshAll(repo);
+        }).catch(function(err) {
+          console.log(err);
+          updateModalText("Push Failed : "+err.message);
```
mehrler12: The error message you get if you aren't signed in is unclear "Push Failed : Method push has thrown an error.".  I see this is the error message coming from nodegit so I'm not sure if there is an easy way to fix this?

Unimax: right. but if you are not signed it, you are not supposed to push or pull right?

Unimax:
merged for now.

PR #84, opened 2019-06-02
Feature 54 view tags on existing commits

hdauvic:
Tags show as separate nodes pointing to the commit they are associated with. Feature supports both annotated tags, and lightweight tags. Additional wiki entry created to explain the graphing.ts file, and the changes made to it to support the tagging graphical interface. The nodes function in the same manner as the branches, and commit nodes in that they can be dragged around freely. When the graph is zoomed in to the fullest extent, a list of tags for each commit is listed along side the Author and Commit Message of individual commits.


SiRumCz:
As I was finishing up my issue #66 before started to collaborate on this issue, I accidentally merged my live-reload implementation to this branch. However this two issues were not correlated and this PR should work fine.

sterlinglaird:
It seems like the original code for generating the graph, it will always show the commits that have a branch associated with them in the upper zoom levels and collapse the commits previous to that commit into that single node (there may be additional logic I don't understand though). If possible, we should try to do that with tags as well so that in the upper zoom levels tags appear associated with commits using the same logic as branches.

hdauvic:
Hmmm... Are they not doing that now @sterlinglaird ? In the repo's I tested it seemed like they were.

sterlinglaird:
Here is an example at the highest zoom level. In both images the `1_commit_back_*` branch/tag is on the same commit which is the commit directly before the most recent commit (where master points to). 

tag
Tag example

branch
Branch example

In the tag example the tag is pointing to a commit inside the collapsed commits, while the branch example points to a separated commit.

sterlinglaird:
If it isn't very difficult, can I also recommend switching the database shape to an ellipse, or just a box with a different color. The database shape looks a little ridiculous with longer tag names haha.

SiRumCz:
> If it isn't very difficult, can I also recommend switching the database shape to an ellipse, or just a box with a different color. The database shape looks a little ridiculous with longer tag names haha.

I would suggest both branch and tag use the same box shape except they have their own icon in the front of the names. Example:
image

Both Icon can be found [here](https://fontawesome.com/icons).

sterlinglaird:
I agree with @SiRumCz his suggestion would be a much better way of differentiating between the two.

hdauvic:
Changing the images *shouldn't* be too difficult. I'll take a look at it.

For the other issue, the way the commits collapse don't allow for a separation the way we would hope. I haven't found a way to do what you're asking it to do at the higher levels.

app/misc/graphing.ts
```diff
@@ -10,15 +10,36 @@ let commitHistory = [];
...
 let spacingX = 80;
+let tagSpacingY = 100;
+let tagSpacingX = 300;
 let parentCount = {};
 let columns: boolean[] = [];
 let edgeDic = {};
 let numOfCommits = 0;
 let branchIds = {};
+let tagIds = {};
+let unumberPrev = 0;
```
colinwerner: This is the previous unique number, correct?

app/misc/graphing.ts
```diff
@@ -10,15 +10,36 @@ let commitHistory = [];
...
+
+// In order to allow tags, branches, and nodes to have unique numerical id's
+// reference: https://stackoverflow.com/questions/8012002/create-a-unique-number-with-javascript-time
+function generateUniqueNumber() {
+    var date = Date.now();
+
+    // If created at same millisecond as previous
+    if (date <= unumberPrev) {
+        date = ++unumberPrev;
```
colinwerner: So if it is <= we increment the value? Logically speaking this makes sense, since it is a date and it should be monotonically increasing; however theoretically, it is a bit confusing. Perhaps update the comment?

app/misc/graphing.ts
```diff
@@ -10,15 +10,36 @@ let commitHistory = [];
...
+    } else {
+        unumberPrev = date;
+    }
+
+    return date;
+}
+
 
 function processGraph(commits: nodegit.Commit[]) {
     var promise = new Promise(function(resolve,reject){
```
colinwerner: Peanut Gallery: What a bizarrely name Class. (I know you didn't create this, but the name perplexes me).

app/misc/graphing.ts
```diff
@@ -257,6 +277,9 @@ function sortBasicGraph() {
         if (idList[i] in branchIds) {
             bsNodes.update({id: branchIds[idList[i]], y: (i + 0.7) * spacingY})
         }
+        if (idList[i] in tagIds) {
+            bsNodes.update({id: tagIds[idList[i]], y: (i + 0.7) * spacingY, x: (i + 0.7) * spacingX})
```
colinwerner: A comment would be nice here.

app/misc/graphing.ts
```diff
@@ -336,7 +362,10 @@ function makeBasicNode(c, column: number) {
 
     if (flag) {
         id = basicNodeId++;
+        tagid = id + 1;
```
colinwerner: Are these supposed to be unique? The line above adds 1, now you're adding another one. Perhaps a comment here?

app/misc/graphing.ts
```diff
@@ -385,11 +415,43 @@ function makeBasicNode(c, column: number) {
...
             });
 
-            branchIds[id] = id + numOfCommits * (i + 1);
+            branchIds[id] = bsnodeId;
+        }
+    }
+
+    // Initializing viewable tags in highest zoom graph level
+    if (c.toString() in tags) {
+        for (let i = 0; i < tags[c.toString()].length; i++) {
```
colinwerner: I'm sure the compiler (or whatever this thing uses) optimizes this code; however, you can cache 'tags[c.toString()].length' as opposed to recalculating it every iteration. I don't know how expensive a.) c.toString() is, b.) the lookup should be constant, and c.) .length could be O(n).

app/misc/graphing.ts
```diff
@@ -385,11 +415,43 @@ function makeBasicNode(c, column: number) {
...
+            let tagName = tags[c.toString()][i];
```
colinwerner: 3rd time to call c.toString() -- cache it?

app/misc/graphing.ts
```diff
@@ -385,11 +415,43 @@ function makeBasicNode(c, column: number) {
...
+            let bsnodeId = generateUniqueNumber();
+            bsNodes.add({
+                id: bsnodeId,
+                shape: "ellipse",
+                // color: "teal",
+                title: tagName, // hover text
+                label: shortTagName, // shown under/in shape
+                physics: false,
+                fixed: false,
+                x: (column - 0.6 * (i + 1)) * tagSpacingX,
```
colinwerner: A comment explaining this math would be nice.

app/misc/graphing.ts
```diff
@@ -453,7 +517,37 @@ function makeAbsNode(c, column: number) {
...
 
                 abEdges.add({
-                    from: id + numOfCommits * (i + 1),
+                    from: bsnodeId,
+                    to: id
+                });
+            }
+        }
+
+        // Initializing viewable tags in second zoom graph level
```
colinwerner: I would suggest indicating what the difference is in this algorithm vs the previous. I understand one is zoomed out. It looks very similar (I do see differences, but perhaps these can be extracted out?)

app/misc/graphing.ts
```diff
@@ -516,7 +624,38 @@ function makeNode(c, column: number) {
...
-                from: id + numOfCommits * (i + 1),
+                from: bsnodeId,
+                to: id
+            });
+        }
+        flag = true;
+    }
+
+    // Initializing viewable tags in lowest graph level
+    if (c.toString() in tags) {
```
colinwerner: This block looks very similar to the previous two?

app/misc/repo.ts
```diff
@@ -258,11 +259,94 @@ function openRepository() {
...
+  Git.Repository.open(repoFullPath)
+    .then(function (repo) {
+      repo.getCurrentBranch()
+        .then(function () {
+          // grab the list
+          return repo.getReferences(Git.Reference.TYPE.LISTALL);
+        })
+        .then(function (refList) {
+          // sort refList alphabetically to get uniform order of the list
+          refList.sort();
```
colinwerner: Since we're doing this every second, I would move the sorting until AFTER the comparison? Sorting is O(nlogn), the comparison is only O(n).

app/misc/repo.ts
```diff
@@ -258,11 +259,94 @@ function openRepository() {
...
+          }
+
+          // detects changes, refresh the lists
+          console.log("branch or tag changes detected... refreshing branch and tag list");
+          bname = {};
+          tags = {};
+          clearBranchAndTagElement();
+          for (let i = 0; i < refList.length; i++) {
+            if (verbose) { console.log("reference name: " + refList[i].name()); }
+            //get simplified name
```
colinwerner: I would like to see this comment expanded, so we can understand how you are parsing the name out.

app/misc/repo.ts
```diff
@@ -258,11 +259,94 @@ function openRepository() {
...
+            let refName = refList[i].name().split("/")[refList[i].name().split("/").length - 1];
+
+            Git.Reference.nameToId(repo, refList[i].name()).then(function (oid) {
+              // Use oid
```
colinwerner: I am good up until this point; however, after here I am a bit lost as to what we are doing.

app/misc/repo.ts
SiRumCz: @colinwerner sorting here is definitely an overhead. However for some reason, the refList here retrived were not always in consistent order, thus there needs a sorting before the comparison. otherwise there is likely a chance it refreshes the list when there is actually no change been made.

app/misc/graphing.ts
SiRumCz: @colinwerner this is the smallest previous unique number.

app/misc/repo.ts
colinwerner: @SiRumCz, thanks for the reply. I misread the code, you're right. The sorting is required.

app/misc/graphing.ts
hdauvic: _ALL_ of the classes are badly named to be honest. They don't entirely make sense. Thus my initiative to write a wiki explaining what's going on.

app/misc/graphing.ts
hdauvic: Unfortunately I'm not entirely sure what this if statement does. It has to exist for everything to work, but it's just copied from the branch behaviour.

app/misc/graphing.ts
hdauvic: I'll comment something in, this may be rendered obsolete from the unique identifiers, but it might be still there for spacing purposes. I'll look into this, thanks!

app/misc/graphing.ts
hdauvic: That's a good thought. They do the same for the branches, however I can add a variable at the beginning of each if statement to calculate this for branches, and tags.

app/misc/graphing.ts
hdauvic: Again, I'm not entirely sure what the math is doing other than giving the node a specific starting spot. Would that be an appropriate comment?

app/misc/graphing.ts
hdauvic: Any suggestions? I am having the hardest time defining the middle ground zoom when it's so similar to the highest zoom level in that the commits are still grouped in a seemingly random way.

app/misc/graphing.ts
colinwerner: Sorry I was referring to to bsNodes.update(...spacingY, spacing X...) that I didn't quite understand.

app/misc/graphing.ts
colinwerner: @SiRumCz the smallest? But it gets updated when date > unumberPrev?

app/misc/graphing.ts
colinwerner: Yep! Just saying what it does conceptually.

app/misc/graphing.ts
colinwerner: I was thinking you could extract this functionality into a helper function, since it is mostly copy/pasted code, something like:
function anAppropriateName(c, tags, nodes, edges) {
    if (c.toString() in tags) {
        for (let i = 0; i < tags[c.toString()].length; i++) {
            let tagName = tags[c.toString()][i];
            let tp = tagName.name().split("/");
            let shortTagName = tp[tp.length - 1];
            console.log(shortTagName + " tag: " + tagName.isHead().toString());
            if (tagName.isHead()) {
                shortTagName = "*" + shortTagName;
            }
            let bsnodeId = generateUniqueNumber();
            nodes.add({
                id: bsnodeId,
                shape: "ellipse",
                // color: "teal",
                title: tagName, // hover text
                label: shortTagName, // shown under/in shape
                physics: false,
                fixed: false,
                x: (column - 0.6 * (i + 1)) * tagSpacingX,
                y: (id - 0.3) * tagSpacingY,
            });

            edges.add({
                from: bsnodeId,
                to: id
            });

            tagIds[tagid] = bsnodeId;
        }
    }
}

Does that make sense?

app/misc/graphing.ts
hdauvic: Id deals with branch array identification, and tagid with tag array identification. If we try to increment basicNodeId again, it fails. But we need tagid to be one ahead of branch id else the whole thing conflicts with itself. basicNodeId++ actually increments basicNodeId at the same time as storing it in id.

colinwerner: Alright, I would suggest adding something to this effect here. Also, consider using ++?

Issue #71, opened 2019-05-29
Pushing or pulling twice will crash application


trstndvll:
This is an extension of issue #30 , where pushing or pulling the application twice in a row will crash the application. We also want to add error handling to the push function as there is currently none, this is being handled in #72 .

trstndvll:
Adding the comments from @Unimax for issue #30:

I found out that the application is crashing because of the 'cred' object we are using for authentication in a callback if failing due to one of the following reason:

it's not able to create the credential object in the C layer. (at least for the 2nd time) .becuase with some changes i was able to see the exact error as in the issue nodegit/nodegit#982
because we are reusing the object it maybe expires after one use and hence having an issue. not sure why.
to prove that is the issue we can simply (I tested on Windows 10 and Ubuntu 18.04)
change the line 406 in master, or 416 in the issue30 branch i.e.
return cred;
into
return Git.Cred.userpassPlaintextNew("username", "password"); 
Note: for now use your username and password. or can use getUsernameTemp() and getPasswordTemp() function from the authenticate.ts

Possible Solutions:

I notice that we are using nodegit instead of Git in some places to generate the cred object. so we can change it to Git. more details on that nodegit/nodegit#982
it will be fixed with the Oauth changes.
On further notes:
Current Flag 'pushButNoCommit' solves an application level optional choice that we want to let the user try to push without any local commits or not. I would say yes (should allow), especially when we don't have any way to detect any local commit done before the start of the application. (by command line or by the same application). Also because with the Auth fixed, it won't crash the application even if you push with no local commits. (tested!!)

moreover, we need to add error handling in case of push actually fails. current flow marks it successful even when it actually fails.

also on linux, i have another issue with the branch name coming as "Branch: master" instead of "master" which I fixed by changing the line 391 (master) . (on Windows this was not a problem)
let branch = document.getElementById("branch-name").innerText.split(": ")[1];

colinwerner:
@Unimax: I saw the "Branch:Master" issue as well, I thought that might be part of the problem, but it didn't resolve anything.

What I was confused by is I see these two commands:
396: addCommand("git push -u origin " + branch);
...
401: return remote.push(

Are we pushing twice?

Unimax:
'Branch: ' got fixed by one who caused it :)
this issue is merged.. Closing.

Unimax:
> addCommand("git push -u origin " + branch);
@colinwerner 
add Command is just to print on screen :)

PR #85, opened 2019-06-03
Feature 53 create tags

sterlinglaird:
Tried to make this pull request earlier but it got deleted for some reason.

This issue is still a work in progress and will need more testing and potentially some design reconsiderations, suggestions are very much welcome!

When in the lowest zoom level (third level), right click on a commit to bring up a newly added context menu which currently only has the option to create a new tag. In the future we can add other functionality to this drop down. It is a little unintuitive to need to be on the lowest zoom level, but that is because in higher zoom levels a node actually represents a collection (or may just be one) of commits while the third level actually represents one commit per node. In the 'Create Tag' menu you can choose the tag name, whether it is an annotated or lightweight tag (and provide an optional tag message), and whether or not to push the tag upstream. After a successful tag creation, the repo is reloaded and the graph is updated to reflect changes that are implemented in #54. Pushing tags multiple times crashes for the same reason as #71 so changes from that may need to be incorporated in this issue.


SiRumCz:
image

Is it better to have the display message here be something like "successfully pushed or created tags xxxxx"?

Adding and pushing lightweight and annotated tag to remote or local repo worked for me, except the app crashed after I played with it couple times.

sterlinglaird:
> image
> 
> Is it better to have the display message here be something like "successfully pushed or created tags xxxxx"?
> 
> Adding and pushing lightweight and annotated tag to remote or local repo worked for me, except the app crashed after I played with it couple times.

Good idea. What was the nature of the crashes? I haven't seen any (at least related to this..) but maybe theres other code paths I didn't think of. The app will crash when pushing multiple times though as identified in #71.

app/misc/git.ts
```diff
@@ -616,6 +616,61 @@ function mergeLocalBranches(element) {
...
+          .then(function (remotes) {
+            return repo.getRemote(remotes[0]);
+          })
+          .then(function(remote){
+            return remote.push(
+              ["refs/tags/" + tagName + ":refs/tags/" + tagName],
+              {
+                callbacks: {
+                  credentials: function () {
+                    return cred;
```
sterlinglaird: This needs to be updated with the new function added with #71

SiRumCz:
Then I suppose I did few pushes. You probably wanna inverstigate on that one.

sterlinglaird:
> Then I suppose I did few pushes. You probably wanna investigate on that one.

An other team is investigating/solving it in #82 /#71. I will incorporate the fix once it merges to master.

Issue #86, opened 2019-06-03
Allow a user to stash local changes


colinwerner:
Decomposed from #34 

This issue is just to allow a user to stash local changes (equivalent of 'git stash' or 'git stash push').

colinwerner:
PR #89 has been submitted to resolve this issue.

PR #89, opened 2019-06-03
Issue 86 allow user to stash changes

colinwerner:
As part of allowing a user to stash/unstash changes, we decided to first implement stashing a change. This change allows a user to create a custom message for the stashed changes.

This PR fixes issue #86.


app/components/header/header.component.html
```diff
@@ -100,10 +100,15 @@
         <li class="toolbar-Icons" onclick="revertCommit()">
             <a href="#"><img src="./assets/Revert.svg" height="40" width="48" class="add-repository-button" title="Revert"></a>
             <span class="toolbar-Text">Revert</span>
-          </li>
+        </li>
+
+        <!--Stash-->
+        <li class="toolbar-Icons" onclick="showStashModal()">
+            <a href="#"><img src="./assets/stash.png" height="40" width="48" class="add-repository-button" title="Stash"></a>
```
mehrler12: Do we have this image in SVG format? I don't see it in font awesome so it would have to be custom. SVG matches the other buttons and keeps it's clarity across resolutions

mehrler12:
Merged in a few fixes from another branch including a fix for #90

app/components/header/header.component.html
colinwerner: No, this was a home made image (originally a JPG) that I converted to PNG with a transparent background and inverted the colours to be consistent with others.

Should we create another issue to make this SVG?

mehrler12: If we don't have one I think it's fine

roweba:
Great job on this, button looks good and the functionality is all there. My only concern is that there is no error handling for when the user is on the "Open a Local Repository Page" and tries to use the stash button. There should be some kind of error that says "Error: Please open a repository to stash changes to".

colinwerner:
> ## Great job on this, button looks good and the functionality is all there. My only concern is that there is no error handling for when the user is on the "Open a Local Repository Page" and tries to use the stash button. There should be some kind of error that says "Error: Please open a repository to stash changes to".

This is a good point. We should do something for all the buttons. git pull has similar behavior.

maxUser:
Just tested it for the first time. Love it!

Issue #81, opened 2019-06-01
signin/signout functionality lost when in hamburger menu


maxUser:
When decreasing the window size so that the hamburger menu appears the sign in/sign out buttons disappear and the functionality is not in the hamburger menu.

maxUser:
I am not able to sign in to visualgit (bug since last sprint ended) so if someone needs to check that the sign out functionality works. Also, if someone can help me with the sign in bug that would be awesome.

PR #93, opened 2019-06-03
Issue 81 hamburgersignin

maxUser:
Added sign in / sign out functionality to the hamburger menu #81 . Since I'm not able to sign in, I haven't been able to test the sign out function. Can someone please do that? Also, there are lots of other things wrong with the hamburger menu.

### DIFF
--- a/app/components/header/header.component.html
+++ b/app/components/header/header.component.html
@@ -2,7 +2,7 @@
   <div class="container-fluid row">
     <div class="navbar-header">
       <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#navbar"
-        aria-expanded="false">
+        aria-expanded="false" onclick="credentialsEntered()">
         <span class="sr-only">Toggle navigation</span>
         <span class="icon-bar"></span>
         <span class="icon-bar"></span>
@@ -103,7 +103,7 @@
           </li>
       </ul>
 
-      
+
       <ul class="navbar-nav navbar-right hidden-xs">
         <li>
           <label id="githubname" style="color:white"></label>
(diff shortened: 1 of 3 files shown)


roweba:
Sign In:
The button is there but the functionality doesn't work. When I sign in after clicking the button it brings me back to the page with the hamburger menu and the button still says "sign in". When I click the sign in button again from this page and enter my credentials, it gets stuck on the "Signing in..." loading screen.
Sign Out:
Everything looks good, no complaints.

app/components/header/header.component.html
```diff
@@ -134,7 +134,8 @@
       </div>
 
       <ul class="nav navbar-nav visible-xs">
-        <li (click)="promptUserToAddRepository()"><a>&nbsp;&nbsp;add repository</a></li>
+        <!-- "add repository" is redundant -->
+        <!-- <li (click)="promptUserToAddRepository()"><a>&nbsp;&nbsp;add repository</a></li> -->
```
nikola-rados: If this is the case should this code just be removed?

app/misc/router.ts
```diff
@@ -58,16 +58,29 @@ function checkSignedIn() {
...
+function credentialsEntered() {
+    // for hamburger menu
+    if (continuedWithoutSignIn) {
+        // not logged in, hide 'sign out' button
+        console.log("user is not logged in");
+        document.getElementById("hamsignout").style.display = "none";
+    } else {
+        // logged in, hide 'sign in' button
+        console.log("user is logged in");
+        document.getElementById("hamsignin").style.display = "none";
```
nikola-rados: I think you could clean this up a touch by factoring out the `document.getElementById...` lines since they are doing the same thing.  The only difference is the variable you are getting, in which case I suggest doing something like this:
```
function credentialsEntered() {
    let status;
    // for hamburger menu
    if (continuedWithoutSignIn) {
        // not logged in, hide 'sign out' button
        console.log("user is not logged in");
        status = "hamsignout";
    } else {
        // logged in, hide 'sign in' button
        console.log("user is logged in");
        status = "hamsignin";
     }
    document.getElementById(status).style.display = "none";
}
```

app/components/header/header.component.html
maxUser: Sure if everyone agrees!

app/misc/router.ts
maxUser: Very nice. I like it.

maxUser:
Alright I'm moving this to sprint 3 because my node/npm is completely non-functional.

colinwerner:
@maxUser I believe we agreed with @jtiz003 to drop this work, correct?

PR #94, opened 2019-06-03
Issue 6 Display Recent Repositories

nikola-rados:
Application will now save all repository path entries and display them in a dropdown menu.  I followed a similar process that the `Remember Login` functionality used.  A `json` file is created/accesses/modified during on every repository input.  At the moment there is a maximum of 5 repositories but the list will remove the oldest entry if that limit is exceeded.
image

This addresses issue #6.


nikola-rados:
Apologies to my reviewers, my IDE automatically fixes what it thinks are unnecessary whitespaces and all those changes are displayed.

app/components/addRepository/add.repository.component.ts
```diff
@@ -44,6 +42,10 @@ export class AddRepositoryComponent {
     }
   }
 
+  getRepos() {
```
Athenasko: Is it possible to add a comment here that give a bit more information of how it's being used?
Say for example, "Allows retrieval of recent repository list in array form for display." Or something. Just for future reference purposes since its a new function.

nikola-rados: Yes, I can definitely do that.

nikola-rados:
@Unimax I'm not quite sure what you mean, would you mind elaborating?

app/misc/repo.ts
```diff
@@ -13,6 +13,79 @@ let span;
...
+        repoList = JSON.parse(checkFile.readFileSync(repoFile));
+    } catch (err) {
+        console.log('Cannot read ' + repoFile);
+        repoList = {
+            recentRepos: []
+        }
+    }
+
+    console.log('Updating recent repos');
+    uodatedRepoList = {
```
sterlinglaird: What is this variable? it doesnt seem to be used. Do we just need to call `updateRecentRepos()`

nikola-rados: Good catch, I forgot to update the `JSON` write.  I wanted to be more explicit about the change such that it was easier to follow the code.  This has been updated, fixed and tested.

nikola-rados:
I wasn't able to find a quick way to add the arrow functionality you wanted @Athenasko.  Is it ok to merge this?

Athenasko:
> I wasn't able to find a quick way to add the arrow functionality you wanted @Athenasko. Is it ok to merge this?

That's alright @nikola-rados. We can set it as a small enhancement task for a future sprint if we really want to.

nikola-rados:
@Athenasko you are right, I have added a commit that takes care of this.

PR #97, opened 2019-06-04
Issue 7 Graph won't update after commit

trstndvll:
This pull request resolves the issue with the graph not updating after committing a change. The problem was found in the `addAndCommit()` function in `app/misc/git.ts`, which was calling unused functions at the end of its run that was blocking the graph update functions from running. We commented out the functions but left them there in case they are needed in future implementations.

Expected behaviour now is once you commit a change, the graph will update and show a new node for the new commit.

### DIFF
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -202,7 +202,7 @@ function addAndCommit() {
       stagedFiles = null;
       hideDiffPanel();
       clearStagedFilesList();
-      clearCommitMessage();
+     // clearCommitMessage();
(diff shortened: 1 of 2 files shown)


trstndvll:
Thanks @colinwerner and @jerushachua ! I'll do another commit taking out those lines. It looks like even though they were commented out in git.ts, when it gets trans-compiled to js it was uncommented.

trstndvll:
Alright I just updated the branch, @jerushachua does it look okay to you? Thanks for your help!

Issue #52, opened 2019-05-22
Allow a user to search tags in current repo


colinwerner:
This is a decomposition of #39.

search current tags in repo (git tag -l "v1.0")

sterlinglaird:
This fix should also fix the problem of branches not being able to search as well, as the mechanism in the drop-down lists should be the same.

PR #98, opened 2019-06-04
Feature 52 - searching tags

jerushachua:
This is a feature implementation for issue #52 . 

I added a function which pulls the input from the search bar and displays only the matching input. The input is case sensitive, since Linux file systems are also case sensitive, but this can be easily changed if needed. 

Please feel free to help test any edge cases I may have missed.


app/misc/git.ts
```diff
@@ -495,6 +495,56 @@ function createBranch() {
...
+                document.getElementById(refName).setAttribute("style","display:block");
+              }
+            }
+
+            lastRefList = refList.slice(); // update lastRefList
+        })
+    });
+}
+
+function displayTag(name, id, onclick) {
```
sterlinglaird: This function already exists in repo.ts?

app/misc/git.ts
```diff
@@ -495,6 +495,56 @@ function createBranch() {
...
+function searchTag() {
+  Git.Repository.open(repoFullPath)
+    .then(function (repo) {
+      repo.getCurrentBranch()
+        .then(function () {
+          // grab the list of references - these could be branches or tags
+          return repo.getReferences(Git.Reference.TYPE.LISTALL);
+        }).then(function (refList) {
+          refList.sort();
+          bname = {};
```
sterlinglaird: Why are these objects getting cleared?

sterlinglaird: And couldnt you just get the tag name from the tags object? that way we dont need to continue querying node git

app/misc/git.ts
jerushachua: Yes - I thought repo.ts is only called when the repo is initialized. If this isn't the case, then I can just remove the duplicate function.

sterlinglaird:
This isn't a requirement for completion of this issue but it wouldn't be very difficult to do:
The search feature would probably be a lot more usable if every time the search bar had input all the tags which have a matching prefix show up. That way if you type in "v1." you'd see "v1.0, v1.0.1, v1.2, v1.3 etc". The "ok" button could also get removed since that is only for creating a branch, which James doesn't actually want to happen through that menu anyways.

app/misc/git.ts
```diff
@@ -495,6 +495,56 @@ function createBranch() {
...
+          tags = {};
+          for (let i = 0; i < refList.length; i++) {
```
sterlinglaird: Theres a missing closing curly bracket for this for loop. My typescript compiler didn't like it

app/misc/git.ts
jerushachua: I wasn't sure if the branches and tags would be reloaded in real time. This only queries nodegit if the tags are being searched for. It's just to ensure we have the latest copy of tags/branches. But you're right that it would be more efficient to use the tags object, if it's up to date. I can test it out with the tags object and if it's working well, I submit a changeset.

sterlinglaird: It probably doesn't matter too much, the object should be up to date though. Regardless, they shouldn't be cleared from within this function, thats probably more important.

SiRumCz:
Based on my experience on old(like ancient) AnuglarJS, I think they have built-in module or mechanics to be bable to filter the tag or branch name by typing partial names. Hope it could help you.

app/misc/git.ts
jerushachua: should be fixed now!

jerushachua:
Fixed all the issues that were mentioned. I removed the OK button and searching is now responsive to user input. Let me know if there are any issues that I missed in my own testing.

app/components/header/header.component.ts
```diff
@@ -1,98 +1,98 @@
...
-  showBranchList(): void {
-    this.branchSelectedInRefDropdown = true
-  }
-
-  showTagList(): void {
-    this.branchSelectedInRefDropdown = false
-  }
-
-}
+import { Component } from "@angular/core";
```
colinwerner: What's with all the whitespace differences?

app/misc/git.ts
```diff
@@ -495,6 +495,37 @@ function createBranch() {
...
+      repo.getCurrentBranch()
+        .then(function () {
+          // grab the list of references - these could be branches or tags
+          return repo.getReferences(Git.Reference.TYPE.LISTALL);
+        }).then(function (refList) {
+          refList.sort();
+          for (let i = 0; i < refList.length; i++) {
+
+            // strip name for readability
+            let refName = refList[i].name().split("/")[refList[i].name().split("/").length - 1];
```
colinwerner: Logically you can move this into the if statement?

Also, can you state what format the name is, please?
ie
name is "/blah/blah/blah/name" and we want to extract "name"?

app/misc/git.ts
```diff
@@ -495,6 +495,37 @@ function createBranch() {
...
+
+            if (refList[i].isTag()){
+              if (refName.indexOf( document.getElementById("tag-name").value ) > -1) {
```
colinwerner: Just to be sure, if you find the refname in the tagname then it returns > -1?

app/misc/git.ts
```diff
@@ -495,6 +495,37 @@ function createBranch() {
...
+                document.getElementById(refName).setAttribute("style","display:block");
+              } else {
+                document.getElementById(refName).setAttribute("style","display:none");
```
colinwerner: lines 515 and 517 could be refactored?
if > -1
  attr = "display:block"
else
  attr = "display:none"
document.getelementbyid(refname).setattribute("style", attr)

app/misc/git.ts
```diff
@@ -810,12 +841,12 @@ function displayModifiedFiles() {
         if (modifiedFiles.length !== 0) {
           if (document.getElementById("modified-files-message") !== null) {
             let filePanelMessage = document.getElementById("modified-files-message");
-            filePanelMessage.parentNode.removeChild(filePanelMessage);
+            filePanelMessage.parentNode.Child(filePanelMessage);
```
colinwerner: I'm just curious about this, what necessitates this change?

app/misc/git.ts
```diff
@@ -495,6 +495,37 @@ function createBranch() {
...
+// search for tags
+function searchTag() {
+  Git.Repository.open(repoFullPath)
+    .then(function (repo) {
+      repo.getCurrentBranch()
+        .then(function () {
+          // grab the list of references - these could be branches or tags
+          return repo.getReferences(Git.Reference.TYPE.LISTALL);
+        }).then(function (refList) {
+          refList.sort();
```
colinwerner: Is the sort necessary if we're doing an indexed search on the name?

app/misc/git.ts
```diff
@@ -495,6 +495,37 @@ function createBranch() {
...
+            let refName = refList[i].name().split("/")[refList[i].name().split("/").length - 1];
+
+            if (refList[i].isTag()){
+              if (refName.indexOf( document.getElementById("tag-name").value ) > -1) {
+                document.getElementById(refName).setAttribute("style","display:block");
+              } else {
+                document.getElementById(refName).setAttribute("style","display:none");
+              }
+            }
+            lastRefList = refList.slice(); // update lastRefList
```
colinwerner: Does slice just return a copy of the array? What exactly is lastRefList used for?

app/misc/git.ts
jerushachua: refName is the readable, string format of the tag name. The string returned from nodegit needs to be stripped before it can be used. 

Which if statement are you referring to?

app/misc/git.ts
jerushachua: Yes, js has a built-in function which helps you compare strings. This will check if refName is a substring of the tag we're currently comparing to.

app/components/header/header.component.ts
jerushachua: I don't know why git detects this as a diff.

app/misc/git.ts
jerushachua: I changed this to test during the functionality of removing nodes. There wasn't any effect by the change here, but I forgot to change it back. I'll change it right now and update the files.

app/misc/git.ts
jerushachua: This was used in the original cleaning of the array when we first grab the tags. I can remove it as it's not needed here.

app/misc/git.ts
jerushachua: Sure, done.

app/misc/git.ts
colinwerner: It's an optimization. You are doing the parsing of the name BEFORE checking if it is a tag. You should check if it is a tag before parsing the name. Hence move let refName =... into the if (refList[i].isTag()) {

app/components/header/header.component.ts
colinwerner: Just revert it.

PR #99, opened 2019-06-04
Issue 70 You cannot see un-pushed commits

trstndvll:
**What did we do?**
This PR adds functionality to allow the user to view the total number of un-pushed commits. We calculate the current number of local commits and the current number of remote commits using nodegit calls, and then take the difference. We have done this by adding 4 new functions in the app/misc/git.ts file:

1. calcUnpushedCommits()
2. unpushedCommitsModal()
3. countLocalCommits()
4. getAllPushedCommits()

We also added a new button called “info” to the header of the app, with its own unique button icon. This triggers the calculation.
 
**Why did we do it this way?**
Originally, we wanted to update the graph and show the un-pushed commits by changing the colour of the nodes. Unfortunately, nodegit does not currently have any functionality similar to that of git log, so we had to divide our own way of counting the commits. We cannot distinguish the un-pushed commits by IDs using nodegit, so the next best thing would be to show the user the count of un-pushed commits. Issues for adding log-type support to nodegit are currently open. 

**Things to note:**
- We are using an asynchronous function call to find the number of remote commits. With Angular 2, there are limitations/issues with async functions. This means that the user must click the “info” button twice in order to get a proper reading from nodegit. This is explained to the user when they hover over the button. 
- This functionality does not affect any other functions or operations within the app.


aeogunmuyiwa:
@colinwerner  Unfortunately we can't,  It is a bit more complicated than calling it twice. The variable does not update globally when we do that because of an ecmascript issue with java script and our version of angular. when we calculate the number of commits the function is nested and the promise first delivers a 0 variable before the actual variable is calculated .

app/components/header/header.component.html
```diff
@@ -427,29 +433,3 @@ <h4 class="modal-title">Confirmation</h4>
   </div>
 </div>
 
-<!--Modal for adding stash message -->
```
colinwerner: You probably don't want to delete this!!!

app/misc/git.ts
```diff
@@ -419,6 +422,35 @@ function pushToRemote() {
     });
 }
 
+//Takes the number of local commits and the number of remote commits and returns the difference
+//This will be the total number of unpushed commits
+@Injectable()
+export function calcUnpushedCommits() {
+  var calc = 0;
+
+  for (let i = 0 ; i < 5; i++){
```
colinwerner: I'm a bit lost with this for loop. Why are we doing it 5 times?

app/components/header/header.component.html
aeogunmuyiwa: It created a merge conflict so I removed it initially , I believe it was not the issue-70 branch but somehow got included during a pull update. However I have updated the file and re-introduced the code.

app/misc/git.ts
aeogunmuyiwa: We made it 5-times just to make sure that the variable is check and updated properly before we use it. It could be twice or one, It was an implementation decision because of the issues we had with calculating the number of unpushed commits locally and the number of commits .

Unimax:
testing failed :( 
Showing me 23 unpushed commits!! which is not true. (i had 1 but even after pushing it showing 24!)

app/misc/git.ts
```diff
@@ -1343,3 +1375,79 @@ function fetchFromOrigin() {
...
+  var aclist = [];
+
+  Git.Repository.open(repoFullPath)
+      .then(function (repo) {
+        repos = repo;
+        return repo.getReferences(Git.Reference.TYPE.LISTALL)
+      })
+      .then(function (refs) {
+        var count = 0;
+        async.whilst(function () {
```
mehrler12: I don't know if you've rebased to master lately but I updated the Async library which changes whilst a little. Take a look at the getallcommits function in master to see the new syntax

aeogunmuyiwa:
- I have included a callback function to the countLocalCommits and getAllPushedCommits function. The variables are updated on function call .
- I have also removed the info button and when an repo is opened on VisualGit start the number of unpushed commit is displayed in a popup Modal,

app/misc/git.ts
aeogunmuyiwa: thanks I have rebased with master and the function is updated now.

Issue #19, opened 2019-05-21
Use Oauth instead of storing the username and password in data.json


omazhary:
Saving the token from Github's Oauth would be more secure than storing a encrypted password in data.json.

Athenasko:
Based on a preliminary search, this feature should require the secure storing of Oauths, and the deletion of the old passwords and methods of storing it.
(https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#non-web-application-flow and https://github.blog/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/)

colinwerner:
I suggest we have a research task this sprint to research this issue and push this to Sprint 2.

norrisng:
@dragonprevost and I have agreed that we're no longer using separate issues as subtasks to this parent issue. Details will be documented in the comments here as necessary.

There's too many little details peppered everywhere (e.g. use of username/password fields) that makes it hard to define a subtask without sounding overly miniscule/specific.

PR #100, opened 2019-06-04
Issue 19: OAuth implementation

norrisng:
This PR implements #19.

Login via username/password has been completely removed, and has been replaced with signing in via GitHub from inside VisualGit:
*  Instead of the username/password, `data.json` now contains the exchanged OAuth token . This token does not expire.
* Login is now always remembered - if there's a `data.json` file containing a valid token, login is automatic upon launching VisualGit.
* Currently, the user will be automatically "signed back in" after clicking "sign out" - OAuth doesn't use the concept of login "sessions", so a "sign out" button may not make much sense here.

The client ID and secret must be obtained from GitHub by registering at https://github.com/settings/applications/new. Currently, this is stored inside `app/misc/authenticate.ts`, under `OauthConfig`. You'll have to supply your own client ID and secret.


app/components/authenticate/authenticate.component.ts
```diff
@@ -9,12 +9,8 @@ export class AuthenticateComponent implements OnInit {
...
-      document.getElementById("rememberLogin").checked = true;
-    } else {
-      // @ts-ignore
-      document.getElementById("rememberLogin").checked = false;
-    }
+      console.log('Logging in With Saved Token');
+    } 
   }
 
   switchToMainPanel(): void {
```
dragonprevost: User credential input checks are no longer required with the new Oauth login

app/components/authenticate/authenticate.component.ts
```diff
@@ -41,52 +37,4 @@ export class AuthenticateComponent implements OnInit {
   authenticateGithub(): void {
     authenticateUser(switchToAddRepositoryPanel);
   }
```
dragonprevost: Oauth login page offers the user to make a new account and reset password etc...

app/components/pullRequestPanel/pull.request.panel.component.ts
```diff
@@ -502,7 +502,8 @@ export class PullRequestPanelComponent {
       url: url,
       type: "GET",
       beforeSend: function (xhr) {
```
dragonprevost: Authorization headers now use access tokens instead of username and password hash

app/components/header/header.component.ts
```diff
@@ -47,44 +47,9 @@ export class HeaderComponent   {
...
-    rid of the warning icon.
-  */
-  notEmptyPassword(): void {
-    if (document.getElementById('Password1').classList.contains('error')) {
-      document.getElementById('Password1').classList.remove('error');
-      document.getElementById('password-error-icon').style.visibility = 'hidden';
-    }
+  authenticateGithub(): void {
+    authenticateUser(switchToAddRepositoryPanel);
+    hideSignInButton();
```
jerushachua: Is this already called in authenticateUser()?

app/misc/readCredentials.ts
```diff
@@ -4,32 +4,40 @@ var jsonfile = require("jsonfile");
...
 
-function getPassword() {
-  var decryptedPasswordBytes = CryptoJS.AES.decrypt(
-    encryptedPassword.toString(),
+  var decryptedTokenBytes = CryptoJS.AES.decrypt(
+    encryptedToken.toString(),
     os.hostname()
   );
-  return decryptedPasswordBytes.toString(CryptoJS.enc.Utf8);
+  return decryptedTokenBytes.toString(CryptoJS.enc.Utf8);
```
jerushachua: Are you encrypting the Github access token here? Is there a need to do this, or is this just an extra security precaution?

dragonprevost: We set a global variable called `encryptedToken` to an encryption object. This section of the code is to decrypt the token and return it as a regular string. The project previously was encrypting the username and password in memory so I just maintained that pattern.

app/misc/authenticate.ts
```diff
@@ -18,15 +19,14 @@ let client;
...
 let signedAfter = false;
-let loginScopes = [
-  "repo",
-  "user"
-];
+let loginScopes = "user repo";
 
+// Configuration of Oauth Application Variables
 const OauthConfig = {
   clientId: 'CLIENT_ID',
```
colinwerner: Can we make this env variables? What happens if they are not set?

app/misc/authenticate.ts
```diff
@@ -84,13 +85,33 @@ function ModalSignIn(callback) {
...
 
 
 function loginWithSaved(callback) {
-    document.getElementById("username").value = getUsername();
-    document.getElementById("password").value = getPassword(); //get decrypted username n password
+  // Get the saved access token from the file system
+  var accessToken = getToken();
+
+  // Store the token in memory
+  encryptTemp(accessToken);
```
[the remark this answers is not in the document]
colinwerner: bump

app/misc/readCredentials.ts
```diff
@@ -4,32 +4,51 @@ var jsonfile = require("jsonfile");
...
+}
+
+// Removes the data.json file from the filesystem
+function deleteToken() {
```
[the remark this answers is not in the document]
colinwerner: bump

app/misc/authenticate.ts
colinwerner: Please let us know what the plan going forward is for the client ID / secret. (Creating a new issue and linking it as a TODO is perfectly acceptable :).

app/misc/storeCredentials.ts
```diff
@@ -31,50 +18,25 @@ function encryptAccessToken(accessToken) {
...
-    var decryptedUsernameBytes = CryptoJS.AES.decrypt(encryptedUsername.toString(), os.hostname());
-    return decryptedUsernameBytes.toString(CryptoJS.enc.Utf8);
-  }
+// Takes in the un-encrypted access token, then encrypts it and stores it in memory
+function encryptTemp(token) {
+  encryptedToken = CryptoJS.AES.encrypt(token, os.hostname());
 }
 
+// Returns the password used for Oauth Credentials
 function getPasswordTemp() {
```
colinwerner: since this is only called from authenticate.ts, I do not see the value of it being in this file -- especially since it returns a hard coded string. Any reason not to move this as a constant to authenticate.ts?

app/misc/storeCredentials.ts
```diff
@@ -4,25 +4,12 @@ var jsonfile = require('jsonfile');
 var fs = require('fs');
 var encryptedPassword;
```
colinwerner: No longer used?

app/misc/storeCredentials.ts
```diff
@@ -4,25 +4,12 @@ var jsonfile = require('jsonfile');
...
 var encryptedUsername;
```
colinwerner: No longer used?

app/misc/storeCredentials.ts
```diff
@@ -4,25 +4,12 @@ var jsonfile = require('jsonfile');
 var fs = require('fs');
```
colinwerner: No longer used?

app/misc/authenticate.ts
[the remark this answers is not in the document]
dragonprevost: The reason we use a function outside of the authenticate file is really only because that is the pattern that exists in the project prior to these changes. This is how encryptedUsername and encryptedPassword were set. I will can change it though.

app/misc/readCredentials.ts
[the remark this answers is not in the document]
dragonprevost: Good idea

app/misc/authenticate.ts
dragonprevost: This is an issue I have discussed with James, he has agreed to address this after the demo tomorrow.

app/misc/storeCredentials.ts
dragonprevost: Once again, since the password is used amongst multiple files, I am hesitant to reference a const directly which has been declared in another file but I understand your point.

Issue #101, opened 2019-06-05
Switching branches/reloading repo sometimes breaks commit graph


sterlinglaird:
To Reproduce:

1. Open a repo with multiple branches
2. Switch branches to a different branch than originally checked out
Result: There are no branches on the graph and appear to be replaced with white circles.

This may not always happen in every repository, so it may be hard to track down, but in my (and my team) experience it is quite common.

![1](https://user-images.githubusercontent.com/13267819/58926477-4cd09a80-8700-11e9-8e8d-e8e2c95b1ee4.PNG)
My sample graph upon opening the repo

![2](https://user-images.githubusercontent.com/13267819/58926489-5823c600-8700-11e9-9b54-8cbb680fa419.PNG)
Graph (at the same zoom level) after switching branches

![3-zoom](https://user-images.githubusercontent.com/13267819/58926636-be104d80-8700-11e9-825a-807ce2817065.PNG)
After zooming in to the second level. There should be at least 2 nodes here but they are all collapsed as one

![4-full-zoom](https://user-images.githubusercontent.com/13267819/58926660-d2ece100-8700-11e9-946f-908d1f0c74ed.PNG)
After zooming to the third (final) level. All 4 commits are displayed correctly but with missing branches.

Issue #29, opened 2019-05-21
Investigate white nodes


omazhary:
After making a commit, white nodes appear on the graph next to the node containing the commit. These white nodes don’t serve any discernible purpose and should be either removed or better labelled.

![image](https://user-images.githubusercontent.com/22312351/57663561-56eff500-75a9-11e9-9b8c-8a3bb6cef969.png)

sterlinglaird:
Related to #101. The issue seems to have gotten more severe since this issue was made since the white nodes alter the graph itself instead of just appearing randomly

PR #116, opened 2019-06-08
ISSUE#105: Allow user to refresh commit graph when detects local repo changes

SiRumCz:
Feature:
- Added a button to allow user refresh commit graph at any time.
- Commit graph can auto detects changes in number of commits in the repo and alert user.

Fix:
- This PR also fixed problem where when at second level zoom, tags and branches are missing in non-HEAD commit.

Design logic:
- From user's point of view, when commit graph detects any changes(whether new tag or branch has been created or new commits has been made), a yellow warning banner will show up with button to allow user to refresh the graph.
- How I achieved this: 1) periodic checks on tag and branch list changes and changes in number of commit current branch has. 2) once detects change, it pops up a banner to tell user that the commit graph is outdated and could be refreshed by pressing the refresh button.

How to test/review this feature:
1. open a repo
2. create or delete tag/branch/commit using command line tools, the refresh button will show up
3. click 'Refresh Graph' button to refresh the commit graph 

Existing problems:
- White node problem might appear after refresh the graph, this problem is under investigation in issue #101 and issue #29. 
- When new commits has been created in current branch, after refreshed the graph, the current branch tag still points to old HEAD commit instead of the new HEAD commit just created. This problem is not caused by this PR because the refresh button only uses existing `drawGraph()` function. I am not sure if this problem should stop me from creating this PR. However, if I have time I might investigate on this one.

### DIFF
--- a/app/components/graphPanel/graph.panel.component.html
+++ b/app/components/graphPanel/graph.panel.component.html
@@ -1,8 +1,9 @@
 <div class="graph-panel" id="graph-panel">
   <!-- auto-detection box on commit graph -->
+  <button id="refresh-button" type="button" class="btn btn-info refresh-graph-btn" (click)="refreshGraph()">Refresh Graph</button>
   <div id="refresh-graph-alert" class="alert alert-warning" role="alert" tabindex="-1">
     Commit graph changes detected!
-    <button id="refresh-graph" type="button" class="btn btn-warning" (click)="refreshGraph()">Refresh Graph</button>
+    <button type="button" class="btn btn-warning refresh-graph-btn" (click)="refreshGraph()">Refresh Graph</button>
   </div>
 
   <div class="network" id="my-network">
--- a/app/components/graphPanel/graph.panel.component.html
+++ b/app/components/graphPanel/graph.panel.component.html
@@ -1,4 +1,10 @@
 <div class="graph-panel" id="graph-panel">
+  <!-- auto-detection box on commit graph -->
+  <div id="refresh-graph-alert" class="alert alert-warning" role="alert" tabindex="-1">
+    Commit graph changes detected!
+    <button id="refresh-graph" type="button" class="btn btn-warning" (click)="refreshGraph()">Refresh Graph</button>
+  </div>
+
   <div class="network" id="my-network">
   </div>
 
--- a/app/components/graphPanel/graph.panel.component.ts
+++ b/app/components/graphPanel/graph.panel.component.ts
@@ -59,7 +59,7 @@ export class GraphPanelComponent {
     createTag(tagName, getSelectedCommit(), pushTag, tagMessage);
   }
 
-  // Initializes  the create tag modal. Should be called before displaying the modal.
+  // Initializes the create tag modal. Should be called before displaying the modal.
   setCreateTagModal() {
     $("#createTagModalCreateButton")[0].disabled = false;
 
@@ -72,9 +72,15 @@ export class GraphPanelComponent {
     let inputPushTag = $("#inputPushTag")[0];
     inputPushTag.checked = inputPushTag.defaultChecked;
     inputPushTag.disabled = signed ? false : true;
-    inputPushTag.title = signed ? "" : "Must be signed in to push tags"
+    inputPushTag.title = signed ? "" : "Must be signed in to push tags";
 
     let createTagError = $("#createTagError")[0];
     createTagError.innerHTML = "";
+  }
+
+  refreshGraph(): void {
+    $("#refresh-graph-alert").hide();
+    drawGraph();
+    updateModalText("Graph successfully refreshed");
   }
 }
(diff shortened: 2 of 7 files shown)


index.html
```diff
@@ -93,15 +93,16 @@
 
   }, 2000);
 
-  // Update modified files on left file panel every X seconds
+  // group all monitors here, call every 3 seconds
```
jerushachua: Is 3 seconds chosen arbitrarily as a value?

app/components/graphPanel/graph.panel.component.html
```diff
@@ -1,4 +1,10 @@
 <div class="graph-panel" id="graph-panel">
+  <!-- auto-detection box on commit graph -->
+  <div id="refresh-graph-alert" class="alert alert-warning" role="alert" tabindex="-1">
+    Commit graph changes detected!
+    <button id="refresh-graph" type="button" class="btn btn-warning" (click)="refreshGraph()">Refresh Graph</button>
+  </div>
+
```
jerushachua: If the user clicks away from the modal without refreshing the graph, can the user refresh it another way?

SiRumCz: The button and alert box won't disappear until user refresh the graph.

index.html
SiRumCz: Original VisualGit chose 3 seconds (3000 ms) to refresh their staged and unstaged file list. I then put my branch/tag live-reload and commit graph change detection into same place.

app/misc/graphing.ts
```diff
@@ -475,6 +475,7 @@ function makeAbsNode(c, column: number) {
     let email = stringer.split("%")[1];
     let flag = true;
     let count = 1;
+    let nodeId;
```
sterlinglaird: I see this code is actually on the branch for issue 54, is this code from previous work that never got merged?

app/misc/git.ts
```diff
@@ -256,6 +256,28 @@ function clearCommitMessage() {
   document.getElementById('commit-message-input').value = "";
 }
 
+// checking if the length of commits is different
+function checkCommitChange() {
```
sterlinglaird: This check isn't complete as the graph can change while the # of commits may stay the same. I don't know how it would be possible to truly detect changes in the graph without fully analyzing the structure of the graph. Perhaps it might just be easier to (perhaps in addition to these changes) provide a button to allow the user to refresh manually?

SiRumCz: @sterlinglaird This implementation is indeed very sloppy, I can't think of any solution that could do the full check without using too much resource. I will include a refresh button.

app/misc/graphing.ts
SiRumCz: @sterlinglaird yes, this fixed problem where tag and branch were missing in the second level zoom. It was fixed after our PR was merged..

SiRumCz:
@sterlinglaird I have added a button to allow user refresh graph at any time.

app/components/graphPanel/graph.panel.component.ts
```diff
@@ -72,9 +72,16 @@ export class GraphPanelComponent {
     let inputPushTag = $("#inputPushTag")[0];
     inputPushTag.checked = inputPushTag.defaultChecked;
     inputPushTag.disabled = signed ? false : true;
-    inputPushTag.title = signed ? "" : "Must be signed in to push tags"
+    inputPushTag.title = signed ? "" : "Must be signed in to push tags";
```
nikola-rados: Nice catch!

app/misc/graphing.ts
```diff
@@ -583,11 +525,70 @@ function makeAbsNode(c, column: number) {
...
+            let shortTagName = tp[tp.length - 1];
+            console.log(shortTagName + " tag: " + tagName.isHead().toString());
+            if (tagName.isHead()) {
+                shortTagName = "*" + shortTagName;
+            }
+            let bsnodeId = generateUniqueNumber();
+            abNodes.add({
+                id: bsnodeId,
+                shape: "ellipse",
+                // color: "teal",
```
nikola-rados: Should this line just be removed?

app/misc/graphing.ts
```diff
@@ -583,11 +525,70 @@ function makeAbsNode(c, column: number) {
...
+                to: nodeId
+            });
+        }
+    }
+
+    // link the tag to updated absNode
+    if (c.toString() in tags) {
+        for (let i = 0; i < tags[c.toString()].length; i++) {
+            let tagName = tags[c.toString()][i];
+            let tp = tagName.name().split("/");
```
nikola-rados: The only place `tp` gets used is in the line below, I think it is just a stylistic choice but I would recommend just putting the variable definition into one line.

app/misc/graphing.ts
SiRumCz: I believe this was an color option made by one of our group member for viewing tags in commit graph which is in other issues.

Issue #49, opened 2019-05-22
Allow visualGit to rename or move a file


colinwerner:
Allow visualGit to rename or move a file (using git commands)

The idea is to replicate the git move functionality (git mv <options>…​ <args>…)

_git move documentation_: https://git-scm.com/docs/git-mv 

The implementation of this feature is open to discussion

PR #117, opened 2019-06-09
Feature 49: Allow user to rename or move a file within a repository (incl. Simple-Git integration)

ghost:
This feature replicates the Git Move ("git mv") functionality allowing a user to rename or move a file within a repository. It also integrates Simple-Git library into the project. 

- Rename and Move buttons were added to the file detail view panel for staged/unstaged files. Modal dialogs prompt for new filename (rename) or target folder (move). A directory browser dialog allows the user to pick a target folder. 
- The Nodegit library does not support Git Move so Simple-Git was selected and integrated as an alternative. Please run `npm install` before starting the project.

### DIFF
--- a/app/components/diffPanel/diff.panel.component.ts
+++ b/app/components/diffPanel/diff.panel.component.ts
@@ -13,7 +13,6 @@ export class DiffPanelComponent{
   private textEditorInstance: TextEditorComponent = new TextEditorComponent();
 
 
-
   /*
     This function is called when the Rename File button is pressed in the diff panel.
     It renames the file currently open in the diff panel in the file editor
@@ -32,10 +31,15 @@ export class DiffPanelComponent{
   This function is called when the Move File button is pressed in the diff panel.
   It moves the file currently open in the diff panel in the file editor
 */
-  moveFileFromDiff():void {
+  moveFileFromDiff(event?: Event):void {
     // Grabs the name of the file that is currently open in the diff panel
     let currentFilename = document.getElementById("currentFilename").innerHTML;
-    let moveFileToFolder = document.getElementById("moveFileToFolder").value;
+    let moveFileToFolder
+    if(event){
+      moveFileToFolder = event.target.files[0].path
+    }else{
+      moveFileToFolder = document.getElementById("moveFileToFolder").value;
+    }
 
     moveFile(currentFilename,moveFileToFolder);
     $('#move-file-modal').modal('hide');
--- a/app/components/diffPanel/diff.panel.component.ts
+++ b/app/components/diffPanel/diff.panel.component.ts
@@ -22,7 +22,7 @@ export class DiffPanelComponent{
     let currentFilename = document.getElementById("currentFilename").innerHTML;
     let renameFilename = document.getElementById("renameFilename").value;
 
-    moveFile(currentFilename,renameFilename);
+    moveFile(currentFilename,renameFilename,true);
     $('#rename-file-modal').modal('hide');
 
   }
--- a/app/components/diffPanel/diff.panel.component.ts
+++ b/app/components/diffPanel/diff.panel.component.ts
@@ -7,12 +7,49 @@ import { TextEditorComponent } from "../textEditor/text.editor.component";
   templateUrl: 'app/components/diffPanel/diff.panel.component.html'
 })
 
-export class DiffPanelComponent {
+export class DiffPanelComponent{
   // Creating instances for the classes so that their functions can be used
   private footerInstance: FooterComponent = new FooterComponent();
   private textEditorInstance: TextEditorComponent = new TextEditorComponent();
 
-  /* 
+
+
+  /*
+    This function is called when the Rename File button is pressed in the diff panel.
+    It renames the file currently open in the diff panel in the file editor
+  */
+  renameFileFromDiff():void {
+    // Grabs the name of the file that is currently open in the diff panel
+    let currentFilename = document.getElementById("currentFilename").innerHTML;
+    let renameFilename = document.getElementById("renameFilename").value;
+
+    moveFile(currentFilename,renameFilename);
+  }
+
+  /*
+  This function is called when the Move File button is pressed in the diff panel.
+  It moves the file currently open in the diff panel in the file editor
+*/
+  moveFileFromDiff():void {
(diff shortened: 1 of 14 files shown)


jpstoldt:
Addresses open issues #49 and #119

colinwerner (commented):
Do I need to install simple-git? or does npm install handle it?

A few comments from my testing:
1.) the dialog is not cleared when you close it (not sure if this is intentional or not?)
2.) Upon renaming a file the dialog remained open (I could see another dialog popup behind it saying the source file is not found).  (It was a DELETED file). So maybe remove these options for these files?
3.) I had a similar result when trying to edit another file, the modal should be closed on rename? 
4.) It says rename successful, but I don't see it being a success?
5.) The contents of issue-49-rename-or-move-file seem different than this PR. (This might be something weird with my setup though.

I will need to look a bit further...

colinwerner:
Upon deleting my node_modules directory and retesting most of the functionality is working as expected.

1.) Select a file to rename
2.) enter asdf.txt as the name
3.) click close
4.) select a new file to rename
5.) asdf.txt is still there.

It should be reset?

maxUser (approved):
Rename and move worked for me! Great stuff JP

app/components/diffPanel/diff.panel.component.ts
```diff
@@ -7,12 +7,58 @@ import { TextEditorComponent } from "../textEditor/text.editor.component";
...
+  /*
+    This function is called when the Rename File button is pressed in the diff panel.
+    It renames the file currently open in the diff panel in the file editor
+  */
+  renameFileFromDiff():void {
+    // Grabs the name of the file that is currently open in the diff panel
+    let currentFilename = document.getElementById("currentFilename").innerHTML;
+    let renameFilename = document.getElementById("renameFilename").value;
+
+    moveFile(currentFilename,renameFilename,true); // call moveFile method in git.ts
```
colinwerner: A comment indicating what 'true' indicates would be beneficial.

app/misc/git.ts
```diff
@@ -1469,6 +1491,22 @@ function fetchFromOrigin() {
   }
 }
 
+/**
+ * This method implements Git Move to rename or move a given file within a repository using the simple-git library
+ */
 
+function moveFile(filesource:string, filedestination:string, ignoreTest:boolean = false) {
```
colinwerner: A comment about ignoreTest would be beneficial.

colinwerner (approved):
Functionality looks good.

I suggest creating another PR to add this functionality to the file editor (so we can rename/move an unedited file).

app/misc/git.ts
```diff
@@ -1,8 +1,10 @@
 import * as nodegit from "git";
 import NodeGit, { Status } from "nodegit";
+import * as simplegit from 'simple-git/promise';
```
nikola-rados: I don't know how to feel about adding `simple-git` to the project.  I don't like the idea of mixing different libraries together to do the same thing, in this case git functionality.  I understand the necessity for it, as `node-git` did not have what you needed.  I still wonder whether it is wise to introduce a second package.

For the purposes of this project it is probably fine, and it works well so I can't complain!

app/components/diffPanel/diff.panel.component.ts
```diff
@@ -7,12 +7,58 @@ import { TextEditorComponent } from "../textEditor/text.editor.component";
...
+    }
+
+    moveFile(currentFilename,moveFileToFolder); // call moveFile method in git.ts
+    $('#move-file-modal').modal('hide');
+    hideDiffPanel();
+  }
+
+
+  /*
+  This function opens the browse directory dialog that allows the user to select a target directory for a file move
```
nikola-rados: Nice job commenting your functions!

app/components/diffPanel/diff.panel.component.ts
```diff
@@ -7,12 +7,58 @@ import { TextEditorComponent } from "../textEditor/text.editor.component";
...
+  This function is called when the Move File button is pressed in the diff panel.
+  It moves the file currently open in the diff panel in the file editor
+*/
+  moveFileFromDiff(event?: Event):void {
+    // Grabs the name of the file that is currently open in the diff panel
+    let currentFilename = document.getElementById("currentFilename").innerHTML;
+    let moveFileToFolder
+    // if folder selected through directory picker dialog, assign target folder from dialog, otherwise assign from textbox
+    if(event){
+      moveFileToFolder = event.target.files[0].path
```
nikola-rados: Missing `;` again.

app/components/diffPanel/diff.panel.component.ts
```diff
@@ -7,12 +7,58 @@ import { TextEditorComponent } from "../textEditor/text.editor.component";
...
+  }
+
+  /*
+  This function is called when the Move File button is pressed in the diff panel.
+  It moves the file currently open in the diff panel in the file editor
+*/
+  moveFileFromDiff(event?: Event):void {
+    // Grabs the name of the file that is currently open in the diff panel
+    let currentFilename = document.getElementById("currentFilename").innerHTML;
+    let moveFileToFolder
```
nikola-rados: It doesn't seem to make a difference but missing `;` at the end here.

app/components/diffPanel/diff.panel.component.ts
nikola-rados: Either what @colinwerner said, or it may be more clear to just name a variable and give it `true` as a value.

nikola-rados (approved):
There were a few missed `;`'s but it doesn't seem to effect the code.  I'd say everything is good as is, nice job @jpstoldt!

app/misc/git.ts
colinwerner: > I don't know how to feel about adding `simple-git` to the project. I don't like the idea of mixing different libraries together to do the same thing, in this case git functionality. I understand the necessity for it, as `node-git` did not have what you needed. I still wonder whether it is wise to introduce a second package.
> 
> For the purposes of this project it is probably fine, and it works well so I can't complain!

I agree with @nikola-rados; however, simple-git has MORE functionality. So going forward we should STOP using nodegit and use simple-git. In fact, we should probably create tickets to start replacing nodegit with simplegit...

Issue #105, opened 2019-06-05
Allow user to refresh commit graph when detects local repo changes


SiRumCz:
Commit graph can auto detects changes in the repo and allow user to refresh graph manually.

sterlinglaird:
Because we already have multiple live reload checks (for file changes and branch/tag updates), it might be worth combining all of the periodic checks into a single check which updates everything in the GUI that needs to be changed.

SiRumCz:
I combined three cheks here into a single monitor to do 3 seconds a check.
```
// group all monitors here, call every 3 seconds
  setInterval(function() {
    // Update modified files on left file panel
    displayModifiedFiles();
    // monitor changes on branch list and tag list
    refreshList(false);
    // check if there is any changes to either ref list or commits
    checkCommitChange();
  }, 3000);
```

SiRumCz:
After my investigation and implementation, I have two concerns for this feature:
1. User will have a great chance running into white node problem when try to refresh graph. 
2. When new commits have been made, after user refreshed the commit graph, the pointing branch will still be pointed to old commit(instead of new commit that just been made).
The first problem is under investigation by @sterlinglaird for this sprint. But for the second problem I am not sure why it doesn't update pointing commit.

Issue #61, opened 2019-05-25
Allow a user to checkout a tag


sterlinglaird:
In git, tags can be checked out just like branches. We should support this feature.

sterlinglaird:
To make this more interactive we could actually just do a right click, on any commit not and have a checkout button since really you can checkout any commit.

SiRumCz:
@sterlinglaird This is a good idea. But since we are trying to keep our last sprint as lightweight as possible, I suggest you to create a new issue about this feature for future students.

sterlinglaird:
@SiRumCz Yea totally. I just made #191

PR #118, opened 2019-06-09
Feature 56 - delete a tag 

jerushachua:
This feature allows you to delete a tag from the tag list. 

**Implementation** 
- The CSS for implementing a button within the list items for the tags was very ugly, so I implemented the ability to double click to delete tag. 
- Upon a single click, a modal pops up to let the user know that they can double click on a tag to delete it. After this first click, the modal doesn't appear anymore. 
- The changes in the tag list are reflected immediately but the graph is not updated. Issue #105 covers that scope. 

Note: there is a minor bug where you cannot delete tag names with periods in them. This is due to JS being unable to handle function passing of strings with periods in the name. I can open an issue for this for further investigation and am open to suggestions on how to fix it.

### DIFF
--- a/app/misc/repo.ts
+++ b/app/misc/repo.ts
@@ -609,27 +609,6 @@ function refreshList(verbose) {
     ul.appendChild(li);
   }
 
-  // deleting tags
-  function deleteTag() {
-
-    // remove taglist item
-    var tagName = name.getAttribute("id");
-    var tagNode = document.getElementById(tagName);
-    name.parentNode.removeChild(name);
-
-    let repo;
-    Git.Repository.open(repoFullPath)
-      .then(function(repoParam) {
-        repo = repoParam;
-      })
-      .then(function(){
-        return Git.Tag.delete(repo, tagName);
-      }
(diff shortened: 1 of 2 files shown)


SiRumCz:
I tried with just passing the name string and it seemed to work and it also handles tag string with periods.
As for the design, I think in the future we will need to make a checkout tag function by single-clicking tag.  Also I am not sure if it is just me, I misclicked a lot.. I think a delete button is more intuitive.

sterlinglaird:
I agree with the delete button being more intuitive, especially because that is how the branches are deleted, and that there is an issue (#61) for checking out a tag which will likely be implemented by simply clicking once on the tag name (because that is how branches do it).

jerushachua:
I added a button based on your comments. It should work with all tag name types after my rework of the delete tag function. 

Thanks @SiRumCz and @sterlinglaird for your suggestions!

jerushachua:
Refreshing the graph based on local changed should be fixed in https://github.com/uvicgse/project-virtual-team-1/issues/105! This was just merged to master today.

PR #120, opened 2019-06-09
Tags and Branches have different images.

hdauvic:
Added clear comments to full file; please review my comments for spelling errors.
Created clear images to differentiate between tags and branches, as well as removing arrows pointing to commits for tags, and changed their lines to be dashed instead of solid to further differentiate them.

I attempted to create boxes with the tags/branch names and an icon within them, but this proved too difficult to fulfill due to limitations with the html/svg communication within the program itself.

### DIFF
--- a/app/misc/graphing.ts
+++ b/app/misc/graphing.ts
@@ -1,3 +1,19 @@
+/* This file initializes all of the values for the nodes in the graphing panel
+   of VisualGit. It uses dataset variables from graphSetup.ts.
+
+   Types of nodes in the network.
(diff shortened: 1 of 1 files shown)


SiRumCz:
Looks good to me, UI looks very cool and I can immediately tell the difference between branches and tags.

hdauvic:
Changed the lines to match the icon colour, as well as updated it so that the text is visible for all themes. I chose arbitrary (and easily changed) colours that are visible on all themes.

jtiz003:
@hdauvic Which issue(s) does this resolve? Be sure to make it clear in the description.
Thanks!

PR #121, opened 2019-06-10
Bug 29 investigate white nodes

sterlinglaird:
This solves #29 and #101.

White nodes should no longer appear when committing or when refreshing the graph.

### DIFF
--- a/app/misc/graphing.ts
+++ b/app/misc/graphing.ts
@@ -16,8 +16,8 @@ let parentCount = {};
 let columns: boolean[] = [];
 let edgeDic = {};
 let numOfCommits = 0;
-let branchIds = {};
-let tagIds = {};
+let branchIds = [];
+let tagIds = [];
 let unumberPrev = 0;
 let selectedCommit: string;
 
@@ -123,7 +123,8 @@ function populateCommits(oldResult) {
         parentCount = {};
         columns = [];
         branchIds = [];
-
+        tagIds = [];
+        
         // Plot the graph
         for (let i = 0; i < commitHistory.length; i++) {
             let parents: string[] = commitHistory[i].parents();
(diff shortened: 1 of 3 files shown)


app/misc/graphing.ts
```diff
@@ -122,6 +122,7 @@ function populateCommits(oldResult) {
         commitList = [];
         parentCount = {};
         columns = [];
+        branchIds = [];
```
sterlinglaird: Solves #29. Previously this was not cleared out so upon refreshing the graph it would already have existing values. The white node was a pre-existing branch in a previous revision of the graph (before commiting).

app/misc/repo.ts
```diff
@@ -344,7 +344,7 @@ function openRepository() {
   }
 
 // works as a monitor for any change to the reference list
-function refreshList(verbose) {
+function refreshList(verbose, force) {
```
sterlinglaird: Solves #101. When refreshing the graph the branch and tag names get cleared in refreshAll() but if no changes are made to the references in git, then they will not be reloaded. The force flag will ensure the tags/branches are reloaded when the graph is refreshed.

app/misc/repo.ts
```diff
@@ -344,7 +344,7 @@ function openRepository() {
   }
 
 // works as a monitor for any change to the reference list
-function refreshList(verbose) {
+function refreshReferences(verbose, force) {
```
sterlinglaird: Solves #101. When refreshing the graph the branch and tag names get cleared in refreshAll() but if no changes are made to the references in git, then they will not be reloaded in refreshReferences(). The force flag will ensure the tags/branches are reloaded when the graph is refreshed.

SiRumCz:
image

I get white node (sometimes, not always) when open a second repo.

1. I open one repo, play with it, checkout different branches.
2. I then open a second repo, sometimes white nodes appear.

sterlinglaird:
@SiRumCz Thanks I'll look into that

sterlinglaird:
Issue @SiRumCz found had the same cause as the other issue, but for tags not being cleared out instead of branches

PR #123, opened 2019-06-11
Feature 38 drag and drop

Unimax:
New feature: modified files (i.e. unstaged files) and staged files can be dragged and dropped into each other. 

Note: prepareDontMissDND function not written correctly and not getting used and hence commented.


Unimax:
> Drag and drop functionality works. Dragging to the "staged files" is very particular; it seems the area is very small.

when did you test it? i have updated it to be 30% which should be enough. and its getting highlighted as well as far as i know.

maxUser:
> > Drag and drop functionality works. Dragging to the "staged files" is very particular; it seems the area is very small.
> 
> @maxUser when did you test it? i have updated it to be 30% which should be enough. and its getting highlighted as well as far as i know.
I just tested it. Might just be a matter of preference.

maxUser:
> > Drag and drop functionality works. Dragging to the "staged files" is very particular; it seems the area is very small.
> 
> @maxUser when did you test it? i have updated it to be 30% which should be enough. and its getting highlighted as well as far as i know.

@Unimax I just tested it. I'm probably just being picky.

app/components/addRepository/add.repository.component.ts
```diff
@@ -69,16 +69,16 @@ export class AddRepositoryComponent {
   }
   prepareDontMissDND :  function() {
 
-      $(document.body).bind("dragover", function(e) {
-          e.preventDefault();
-          return false;
-      });
+      // $(document.body).bind("dragover", function(e) {
```
colinwerner: Suggestion, either:
1.) delete these
2.) provide commentary as to why they are commented out.

Unimax: Oki i did write in in the PR . but you are right will write it in code also.

Unimax:
> Sometimes the drag and drop animation shows the file going back to the original list (not sure what the steps are to repro). (Perhaps a future TODO?)
> 
> It seems weird that for a moved file the "deleted" file and the "new" file are NOT linked. Perhaps this is something that @jpstoldt could like at. Either way, I think we should create a new issue for this for sprint 4 to link these two items together, somehow?
> 
> There are a lot of UI changes in the TS code -- is this acceptable? I do not know; but there are a lot of hard coded things that might lead to maintainability issues in the future -- hopefully not within the next 3 weeks).
> 
> @mehrler12 @maxUser we need to work with @Unimax to manage this merge with #124 -- (ie we need to test how these changes interoperate with each other. Hopefully, no issues arise!). Probably create a branch with both changes to give it a quick test?

I think if we are considering maintainability issues we should start with our angular version which I can do to highest of version 2. i.e. 2.4.10 I think.  this will let further integration with many plugins possible without issue.
This kind of dynamic elements generation in a js I have seen before. I don't know what else is the correct way. certainly it could be improved.
I was not able to see the UI issue you mentioned.

Unimax:
merged for now.

Issue #88, opened 2019-06-03
Allow a user to pop a stash 


colinwerner:
Decomposed from #34 

Allow a user to 'git stash pop' or 'git stash pop --index').

Issue #87, opened 2019-06-03
Allow a user to view a list of stashes


colinwerner:
Decomposed from #34 

This issue tracks the ability to allow a user to view a list of stashes (the equivalent of 'git stash list'). (Design consideration should also be taken into consideration to allow a user to 'git stash pop' or 'git stash pop --index').

PR #124, opened 2019-06-11
Issue 87 allow a user to view stash list

mehrler12:
#87 Stash list displays and live updates with all current stashes. Should update similarly to the staged files list. Also allows for popping the top stash in the list as per #88 , pop button should hide on if there are no stashes to pop.

### DIFF
--- a/app/components/filePanel/file.panel.component.html
+++ b/app/components/filePanel/file.panel.component.html
@@ -32,8 +32,7 @@
   </div>
   <div class="files-stashed" id="files-stashed">
     <p class="stashed-files-message" id="stashed-files-message">Your stashed files will appear here</p>
-    <div class="file" *ngFor="let file of files">
-      <p>{{file}}</p>
-    </div>
+    <ul id="stash-list" class="list-group">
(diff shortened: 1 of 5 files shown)


Athenasko:
image
Is it possible to make the message more clear? Or if it not possible to determine more specific information based on the error thrown?

Other than that, everything is looking good so far.

colinwerner:
> image
> Is it possible to make the message more clear? Or if it not possible to determine more specific information based on the error thrown?
> 
> Other than that, everything is looking good so far.

Hi @Athenasko, what exactly did you do? I presume you tried to pop and there was a merge conflict, if so, I'd rather not add custom messages in this change. We can certainly look to add them at a later time.

Athenasko:
> Hi @Athenasko, what exactly did you do? I presume you tried to pop and there was a merge conflict, if so, I'd rather not add custom messages in this change. We can certainly look to add them at a later time.

The error came up as I was attempting to pop a stash with one untracked and one tracked file. Though, yes, there would have been a merge conflict in the tracked file.

PR #131, opened 2019-06-11
Issue -70 - recloned master and updated unpushed commits count

aeogunmuyiwa:
I recloned master and copied the unpushed commits functions, however for the getallpushed commit function I rewrote it a bit different. 

This PR adds functionality to allow the user to view the total number of un-pushed commits. We calculate the current number of local commits and the current number of remote commits using nodegit calls and then take the difference. We have done this by adding new functions in the app/misc/git.ts file.

We cannot distinguish the un-pushed commits by IDs using nodegit, so the next best thing would be to show the user the count of un-pushed commits. Issues for adding log-type support to nodegit are currently open. simple-git was added to the project later into this sprint, so we didn't have time to test out its logging functions for our issue. This may be looked into in a future sprint.
The count is displayed to the user with a counter on the right-hand side of the application, just under the Pull Request button. This will update every 3 seconds as per the other functions in setInterval() in index.html.
This PR is for a new branch, "issue-70-unpushedcommits-pr ," as we were unable to resolve the merge conflicts of our original branch "issue-70-unpushed-commits." We tried to rebase our branch onto the current version of master, but the culmination of the node package updates and the length of time since we created the issue-70 branch resulted in too many issues. You can still view that branch for full commit history.

### DIFF
--- a/app/components/pullRequestPanel/pull.request.panel.component.html
+++ b/app/components/pullRequestPanel/pull.request.panel.component.html
@@ -23,7 +23,14 @@
     <div class="right-pr-nav">
         <i (click)="togglePRPanel()"><img src="./assets/PullRequestIcon.svg"></i>
         <hr>
-        <button class="center-block" height="40" width="48" id ="unpushed"  font-weight="bold" title="number of un-pushed commits"></button>
+        <a class="badge badge-pill badge-light text-left "  style="transform: scale(0.7);">Ahead
+            <span class="badge badge-light"  id ="ahead_count" ></span>
+        </a>
+        <a class="badge  badge-dark btn-responsive text-left" style="transform: scale(0.7);">Behind
+            <span class="badge badge-light"  id ="behind_count" ></span>
+        </a>
+
+<!--        <button class="center-block" height="40" width="48" id ="unpushed"  font-weight="bold" title="number of un-pushed commits"></button>-->
     </div>
 
 
--- a/app/components/pullRequestPanel/pull.request.panel.component.html
+++ b/app/components/pullRequestPanel/pull.request.panel.component.html
@@ -23,7 +23,7 @@
     <div class="right-pr-nav">
         <i (click)="togglePRPanel()"><img src="./assets/PullRequestIcon.svg"></i>
         <hr>
-        <button class="btn btn-primary center-block" height="40" width="48" id ="unpushed" font-weight="bold" title="number of un-pushed commits">  </button>
+        <button class="center-block" height="40" width="48" id ="unpushed"  font-weight="bold" title="number of un-pushed commits"></button>
     </div>
 
 
--- a/app/components/pullRequestPanel/pull.request.panel.component.html
+++ b/app/components/pullRequestPanel/pull.request.panel.component.html
@@ -22,7 +22,10 @@
     <!-- Nav bar on the right of the screen to show/hide PR Icon. -->
     <div class="right-pr-nav">
         <i (click)="togglePRPanel()"><img src="./assets/PullRequestIcon.svg"></i>
+        <hr>
+        <button class="btn btn-primary center-block" height="40" width="48" id ="unpushed" font-weight="bold" title="number of un-pushed commits">  </button>
     </div>
+
 
     <div id="pr-status-2">
         <!-- Area to list all PRs. -->
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -23,6 +23,8 @@ let commitHead = 0;
 let commitID = 0;
 let lastCommitLength;
 let refreshAllFlag = false;
+let total_commit ;
+let commit_diff ;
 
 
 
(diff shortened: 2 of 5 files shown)


aeogunmuyiwa:
This PR adds functionality to allow the user to view the total number of un-pushed commits. We calculate the current number of local commits and the current number of remote commits using nodegit calls, and then take the difference. We have done this by adding 4 new functions in the app/misc/git.ts file:
    calcUnpushedCommits()
    unpushedCommitsModal()
    countLocalCommits()
    getAllPushedCommits()

The unpushedCommitsModal() is called in the SettimeInterval function at index.html and the values are updated every 3000 ms.

aeogunmuyiwa:
https://slack-files.com/TJBDCRM6G-FK4PT3Y4B-a3bdba6669
The link about contains the functionality image. 
- because of lots of update to master, we had issues resolving the merge conflicts as the upgrade to angular and file changes had occurred.
-  I cloned master and copied the unpushed commits functions, the main merge conflicts issue was coming from getAllPushedCommits() function so I rewrote it to match the create changes in the git.ts file.
-I created a new branch called issue-70-unpushedcommits-pr where the current working code can be found.

app/components/pullRequestPanel/pull.request.panel.component.html
```diff
@@ -22,8 +22,11 @@
     <!-- Nav bar on the right of the screen to show/hide PR Icon. -->
     <div class="right-pr-nav">
         <i (click)="togglePRPanel()"><img src="./assets/PullRequestIcon.svg"></i>
+        <hr>
+        <button class="btn btn-primary center-block" height="40" width="48" id ="unpushed" font-weight="bold" title="number of un-pushed commits">  </button>
```
mehrler12: - I don't love this being a button, hovering over it leads me to believe this is clickable which it isn't. Perhaps using a badge might be better https://getbootstrap.com/docs/3.4/components/#badges
- It also isn't immediately obvious what this number means until you hover over it which I suppose is okay, maybe moving it to near the commit panel or the push button might be more clear?

aeogunmuyiwa: Thanks, @mehrler12 A badge also used the button tag that we used as per your reference. A button tag becomes clickable when you include an onclick() function which we don't use. 
-Also, James approved the position and the current UI because the number would affect how the commit panel is displayed.

aeogunmuyiwa: Also when you open a fresh repo, the number is not -1. The number only updates if there are changes on a repo locally that has not been pushed.

mehrler12: image
Okay, no worries on the placement then! 
Badges use the span tag, not the button tag. I realize that the button is only clickable when you set onClick but it still looks like it's clickable without it (Changing the pointer and colour) which is where my concern comes from.

trstndvll:
@aeogunmuyiwa Maybe we could change the background colour of the badge to white so it looks less clickable?

app/components/pullRequestPanel/pull.request.panel.component.html
trstndvll: @aeogunmuyiwa Maybe we could change the background colour of the badge to white so it looks less clickable?

aeogunmuyiwa: @mehrler12 Thanks for that catch, I fixed the variable subtraction. If you could pull the branch and test it again that would be awesome. @trstndvll  I changed the background color to white now. thanks.

mehrler12: Still -1 
image
I'll give you access to my dummy repo, maybe it's something in there?
As per the button I was fine with the colour before, just that it changes on hover but if it's too much to change that's alright how it is! I'll submit my approval

Athenasko: Continuing on from @mehrler12, and with the newest changes, it still acts like a button. While changing the color was a good idea, on mouse over it still 'bezels'/highlights and makes it seem like a button. Is there a way to stop that from happening @aeogunmuyiwa?

aeogunmuyiwa: There’s no such functionality like acts like a button . James wanted it look like a badge and that’s a badge. It’s only a button if it becomes clickable (onclick ) function which it doesn’t have . We used the hover to give it a context . Thanks @Athenasko

trstndvll: > Still -1

@mehrler12 Interesting! I'm on a Mac and I'm not even able to get into a new repo to test this out. I go to make one and it just spins endlessly at "loading graph." It's the same story on current version of Master. Is creating a local repo working as expected for you?

aeogunmuyiwa: @trstndvll I don’t think that issue is related to this ticket though . It’s has something to do with master I suppose

mehrler12: @trstndvll I added you to a private repo I've been using for testing try cloning that?

colinwerner:
I see a -2 in my repo:

commit 9fdd244c40b4a24dba5f3823a703d8fb8b02a9c5 (HEAD -> master, origin/master, origin/HEAD)
Author: Colin Werner <colinwerner@gmail.com>
Date:   Wed Jun 5 20:25:31 2019 -0700
    whatever

...
test-merge $ git push
Everything up-to-date
test-merge $ git pull
Already up to date.

colinwerner:
PS: I think we can use simple-git and it will do everything for us:
let sGitRepo = sGit(repoFullPath);

  sGitRepo.silent(true).status().then((status: StatusSummary) => {
    console.log(status);
  });

image

Unimax:
for me, it's failing to show current local changes correctly.
Steps to recreate:-
-open a repo visual git commit 2 time
-open same repo on remote using GitHub and commit 2 time

Expect result: 2     getting: 0

Testing failed :(

Unimax:
> Approved pending -1 fix

I suggest not to approve in case of a blocking/important bug. As 2 approval signal merge permission.

aeogunmuyiwa:
@Unimax @mehrler12 @trstndvll  Thank you guys for pointing out a major flaw in my implementation. I have rewritten the whole function and it now displays the ahead and behind numbers. Please test it out and I would appreciate your comments. Thanks

aeogunmuyiwa:
All approvals are in and the issue is closed

aeogunmuyiwa:
All approvals are in and the issue is closed

app/misc/git.ts
```diff
@@ -23,6 +23,8 @@ let commitHead = 0;
 let commitID = 0;
 let lastCommitLength;
 let refreshAllFlag = false;
+let total_commit ;
```
colinwerner: If these vars are no longer needed then we can remove them?

Issue #31, opened 2019-05-21
GitHub profile pictures on nodes


omazhary:
Currently, only the first letter of GitHub users' names appears on the nodes. It would be more useful to have their GitHub profile pictures so they can be more easily identified.

norrisng:
Current implementation doesn't work for authors that do not use `@users.noreply.github.com` emails. See #157 for more details.

PR #135, opened 2019-06-12
Issue 31: GitHub profile pictures on nodes

norrisng:
This PR implements #31. 

In addition, GitHub profile pictures are now also available in the popup that appears when clicking on a node:

commit_popup

In order to retrieve a user's profile picture through the GitHub API, the username is required. However, because usernames are a GitHub concept - Git works with full names and emails, as supplied via `git config user.name` and `user.email` - **GitHub usernames are scraped from `@users.noreply.github.com` emails if available**. Consequently, profile pictures are unavailable for users that did not hide their email in their GitHub settings and set their Git `user.email` to their `noreply` email ([more info here](https://help.github.com/en/articles/about-commit-email-addresses)).

Letter icons will continue to be used if a profile picture is unavailable.

There does not appear to be a way around this, as a Git repository simply does not contain the needed info required by the GitHub API. However, the existence of other Git GUIs (e.g. GitKraken) seems to prove otherwise. More investigation on this will be needed.

As well, some refactoring of `misc/images.ts` was also done: 
 * `img4User()` was rolled into `imageForUser()` 
    * `img4User()` is no longer used in this branch, but I've kept it in case anyone else is
 * Retrieval of letter-icon images was moved into a helper function
 * Existing unused/commented-out code was also removed

`misc/graphing.ts` was also slightly modified to accommodate the change to `imageForUser()` (in particular, the use of callbacks).

### DIFF
--- a/app/misc/images.ts
+++ b/app/misc/images.ts
@@ -52,25 +52,4 @@ function imageForUser(name: string, email: string, callback) {
     pic = getLetterIcon(name);
   }  
   callback(pic);
-  // githubAvatarUrl(username, function(err,avatarURL){
-  //   if(!err && !email.includes('@users.noreply.github.com')) {
-  //     console.log(`avatar url: ${avatarURL}`);
-  //     pic = avatarURL;
-  //   } 
-  //   // fallback to letter icons if the email isn't a GitHub noreply one
-  //   else {
-  //     let first = name.trim().charAt(0).toUpperCase();
-  //     pic =  "node_modules/material-letter-icons/dist/png/" + first + ".png";
-  //   }
-    
-  // });
-
-  // if (images[email] === undefined) {
-  //   images[email] = "assets/img/starwars/" + imageFiles[imageCount];
-  //   imageCount++;
-  //   if (imageCount >= imageFiles.length) {
-  //     imageCount = 0;
-  //   }
-  // }
-  // return images[email];
 }
--- a/app/misc/images.ts
+++ b/app/misc/images.ts
@@ -46,6 +46,7 @@ function imageForUser(name: string, email: string, callback) {
     else {
       // github-avatar-url insists on having an API token, 
       // but since we're don't need it for public info, we'll instead use octonode to avoid having said token.
+      // That being said, if you've hit the rate limit, just supply the token below; i.e. gh.client('token')
       let client = gh.client();
       
       client.get(`/users/${username}`, {}, function (err, status, body, headers) {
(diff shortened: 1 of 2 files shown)


trstndvll:
Whoops, sorry for removing my review request! It was an accident. So I set my email to private and also set up my no-reply email on git in terminal and I still wasn't able to see my profile pic on commits. Is there something I'm missing? Thanks.
Screenshot 2019-06-11 21 28 09
Screenshot 2019-06-11 21 29 49

norrisng:
@trstndvll Hmm, looks like the numbers appended to the username can vary in length. [The documentation said there were exactly 7](https://help.github.com/en/articles/about-commit-email-addresses), so I just used that in the `string.replace()` regex.

I'll change it to 4+ digits, I guess.

app/misc/images.ts
```diff
@@ -1,40 +1,74 @@
...
-      pic =  "node_modules/material-letter-icons/dist/png/" + first + ".png";
+  
+  let pic;  
+  
+  if (email.includes('@users.noreply.github.com')) {
+    
+    // Extract GitHub username from noreply email (<number>+<username>@users.noreply.github.com)
+    // This won't work for users that are NOT hiding their email in the GitHub settings.
+    // See https://help.github.com/en/articles/about-commit-email-addresses for more info.
+    let username = email.replace('@users.noreply.github.com','').replace(/^(\d){4,}\+/,'');
```
dragonprevost: You can use account.login(initialized in `authenticate.ts`) to get the usersname.

app/misc/images.ts
```diff
@@ -1,40 +1,74 @@
...
     }
-    callback(pic);
-  });
+    else {
+      // github-avatar-url insists on having an API token, 
+      // but since we're don't need it for public info, we'll instead use octonode to avoid having said token.
+      // That being said, if you've hit the rate limit, just supply the token below; i.e. gh.client('token')
+      let client = gh.client();
+      
+      client.get(`/users/${username}`, {}, function (err, status, body, headers) {
```
dragonprevost: This call is already made in authenticate.ts when the user logs in. You can access the response with the account object

norrisng:
@mehrler12 Thanks for the find! Looks perfect.

That being said, it's a *tad* slow, probably because it uses the GitHub Commit Search API (still in beta), and their backend DB likely doesn't index by email.  Probably a good idea to keep the remaining `noreply`-based implementation, but use `github-username` as a fallback.

Gonna implement it next sprint as #157 (it'll require some refactoring of `images.ts`).

norrisng:
@dragonprevost Yep, that would be #157. Totally doable, but requires some refactoring, so that'll be for the next sprint. Thanks for the approval though!

PR #147, opened 2019-06-12
Work around for the tsc errors on npm start

Unimax:
its a workaround.
i did spend around 2 hours trying to fix tsconfig.file so that these error wont come but not able to reduce to 0. the minimum was 235 errors!

so, for now, added this dummy real error.

### DIFF
--- a/app/components/addRepository/add.repository.component.ts
+++ b/app/components/addRepository/add.repository.component.ts
@@ -67,6 +67,10 @@ export class AddRepositoryComponent {
   returnToMainPanel(): void {
     switchToMainPanel();
   }
+
+//creating dummy error so that tsc warning wont show up
+generateDummyError:  function(){}
+
   // old code, not written correctly and not getting used and hence commented
   // prepareDontMissDND :  function() {
       // $(document.body).bind("dragover", function(e) {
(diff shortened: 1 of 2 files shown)


wliang93:
Dummy error shows up on the git console in Windows. Looks a lot cleaner at the moment. 
Are the warnings showing up because of SimpleGit? If it is have other people fixed the issue before?

Issue 143

Unimax:
> Dummy error shows up on the git console in Windows. Looks a lot cleaner at the moment.
> Are the warnings showing up because of SimpleGit? If it is have other people fixed the issue before?
> 
> Issue 143

I am pretty sure it's not because of simple-git.

Unimax:
the current solution adds an extra error in the console which is not intended, will reopen after fixing it.

dragonprevost:
Hey @Unimax , do we know the source of the console errors as of now? I am a bit confused about where the errors are coming from since they did not exist last sprint.

Unimax:
> Hey @Unimax , do we know the source of the console errors as of now? I am a bit confused about where the errors are coming from since they did not exist last sprint.

they were there just not shown as a real error is given priority. these errors are output by tsc (type script compiler). they are mainly warning which are shown as error based on our tsconfig.js.

Issue #151, opened 2019-06-13
remove repetitive log lines


colinwerner:
These log lines loop over and over in the console.

waiting for repo to be initialised
0
0

PR #152, opened 2019-06-13
comment out a couple console.log lines that were printing too often

colinwerner:
now when you use the console in VG you don't see the log lines repeating


colinwerner:
fixed #151

PR #154, opened 2019-06-13
updated the git signatures to the user who is logged in instead of th…

dragonprevost:
This pull request has a new function to generate Signatures based on the account that has logged in instead of the Oauth application account. Everywhere that `repo.defaultSignature()` is called, has been replaced with `getSignature()`.

### DIFF
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -537,7 +537,10 @@ function getUsername(){
 }
 
 // Get a git signature for the account that is logged in.
-function getSignature(){
+function getSignature(repository){
+  // If the user has deffered loggin in, use the default signature
+  if (!account)
+    return repository.defaultSignature();
   // Set email to the email in account object
   var email = account.email;
   // If the email is null, change it to <uesrname>@users.noreply.github.com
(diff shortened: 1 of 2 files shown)


app/misc/authenticate.ts
```diff
@@ -535,3 +535,14 @@ function getUsername(){
   // Return the users username
   return account.login;
 }
+
+// Get a git signature for the account that is logged in.
+function getSignature(){
+  // Set email to the email in account object
+  var email = account.email;
```
colinwerner: Probably need to check for account (in the case when I am not logged in)?

image

dragonprevost: Good catch, did this happen when you differed logging in?

colinwerner: Yes. "continue without sign in" then stash some changes.

dragonprevost: I updated the get signature function to take the repository as a parameter and return the default signature if the user has not logged in. Give it a shot and let me know if it solves this issue! :)

jerushachua: Is '@users.noreply.github.com' the only type of email that will appear?

dragonprevost: If the account has a public email address, then that public email address will be applied to the signature, if not the email will follow the format of `<uesrname>@users.noreply.github.com`.

PR #161, opened 2019-06-15
Issue 132 - Show Upstream Repository

roweba:
This pull request implements issue #132 . The currently configured upstream repository will be displayed in the upstream modal.


app/misc/git.ts
```diff
@@ -1618,39 +1618,94 @@ function cleanRepo() {
...
-      },
-        function (err) {
-          console.log("Waiting for repo to be initialised");
-          displayModal("Please select a valid repository");
-        });
-  } else {
-    displayModal("No Path Found.")
+      repository = repo;
+      var result = Git.Remote.createWithFetchspec(repository, 'upstream', upstreamRepoPath, '+refs/heads/*:refs/remotes/upstream/*');
+      addCommand("git remote add upstream " + upstreamRepoPath);
```
colinwerner: You are adding the command twice (line 1656 and 1665). You should remove one?

colinwerner: Note this doesn't occur with issue 133.

subahm: I think we should remove the command from line 1656 and leave it at line 1665 after checking all the conditions.

colinwerner: I think the addCommand line should be right before you issue the command to nodegit/simple-git. This way IF an error appears, then the user can see the last command issued and the applicable error. Make sense?

roweba: I actually already solved this in PR #162. I deleted the duplicate command on line 1665 and replaced it with a "success" modal. This way the user knows that the command ran successfully.

roweba:
Gonna merge this as part of #162

Issue #133, opened 2019-06-11
Allow user to edit and delete upstream repository


roweba:

roweba:
Deleting functionality is done. Editing functionality is there, but will be improved in Sprint 5 #172

PR #162, opened 2019-06-16
Issue 132 and 133 - Show and Delete Upstream Repository

roweba:
Issue #132 and #133
This allows the user to delete the currently configured upstream repository. In the "Upstream" modal there is a "Delete" button in the user has an upstream configured. Clicking this removes the current upstream.
I decided not to include an "Edit" button because if the user wants to edit the upstream they can just enter the new url into the upstream modal.


app/misc/git.ts
```diff
@@ -1618,39 +1618,116 @@ function cleanRepo() {
...
+ * Delete the current upstream repo.
+ */
+function showUpstreamDelete() {
+  $('#delete-upstream-modal').modal('show');
+}
+
+/**
+ * Delete the current upstream repo.
+ */
+function deleteUpstream() {
```
colinwerner: If i enter something into the upstream box:
image

Then I git 'delete' then I either 'cancel' or 'confirm' then the status dialog is still filled out.

roweba: The modal should now clear. What do you mean by deleting and setting the upstream in one command?

colinwerner: If I enter a new path and hit confirm it should say "upstream already exists, do you want to replace with blah?" confirm/close ... and then it deletes and updates behind the scene?

app/components/header/header.component.html
```diff
@@ -439,3 +427,50 @@ <h5 type="text" id="stashMsgText" class="stashMsgModalComponents"> Stash Message
...
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+        <h4 class="modal-title">Confirmation</h4>
+      </div>
+      <div class="modal-body">
+        Are you sure you want to delete the upstream repository?
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+        <button type="button" class="btn btn-warning" id="localDeleteButton" data-dismiss="modal" onClick="deleteUpstream()">Delete</button>
```
mehrler12: Should this be `btn-danger` to match the delete button on the previous page?

roweba: Do you have a repository open when it looks like this?

mehrler12: Yeah, If I highlight the text it shows up properly! (Same case with a saved upstream)
image

app/misc/git.ts
roweba: Sure, l'll create an issue and work on it in the next sprint.

app/components/header/header.component.html
roweba: Are you on Windows? Interesting....

app/misc/git.ts
roweba: Issue #172

app/components/header/header.component.html
mehrler12: Yeah I'm on windows! Seems like what's happening is in color.ts it finds every <p> element and colors it depending on the style (for default this is white). Not sure why it isn't happening on Mac, I think you could probably fix it by changing the <p> tag to something else or adding a colour with the !important modifer to the CSS

roweba: I took your suggestion and used the !important tag in the CSS and also changed the button to `btn-danger` , please see the latest update!

mehrler12: Works for me! Nice work!

roweba:
> Yep, everything works on my end!
> 
> That being said, if I don't have a repo open, the upstream modal is still accessible, including the "delete" button in it.
> 
> image
> 
> While attempting to set the upstream repo triggers the expected error message, clicking on "delete" will trigger the deletion confirmation modal, and clicking "delete" there will close that modal (it won't reopen the upstream modal), without triggering the "successfully deleted" modal.
> 
> All that being said, VisualGit doesn't crash or fill the console log with any uncaught errors, so it's just a harmless visual glitch.

Good catch! But honestly, I don't even know why we're even showing the navbar when the user doesn't have a repository open?? I don't think it serves any purpose, as those icons are used to make modifications to a repo.

PR #163, opened 2019-06-16
Feature 141 - Deleting a tag from the graph

jerushachua:
This feature lets you delete tag from the graph. 

You can right click on the commit which contains the tags you would like to delete and click "Delete a tag". The tags which are associated with the commit will show underneath and you can delete the selected tag. The tag is deleted right away and disappears from the list, and the graph will prompt the user to refresh the graph, which will update to show the correct changes.


sterlinglaird:
Deleting a tag works good but I have a few comments:

1. When I click the delete button on a node and then click away from the node to close the context menu, and then right click again on a node the list of commits is still open. That list should probably always be closed every time I right click on a node.
2. Its unintuitive to me that I can right click on any node and delete any tag from that node. The list of tags to delete (in my opinion) should either only contain tags for that specific node, or the user just right clicks on the tag icon itself and selects the 'delete' option there.

SiRumCz:
deleting function works, but I would suggest user can right click each tag to delete them.

norrisng:
Yep, feature works as advertised. 

That being said, while the "graph change" message does imply successful tag deletion, it's somewhat unintuitive. Perhaps something like a modal window that says "Deletion successful - please refresh graph to show changes" might be more intuitive?

jerushachua:
> deleting function works, but I would suggest user can right click each tag to delete them.

Unfortunately the DOM structure doesn't reference each node - when right clicking, you can only specify the commit node. I can change the behavior to display the tags associated to each commit instead of all current tags.

jerushachua:
> Deleting a tag works good but I have a few comments:
> 
> 1. When I click the delete button on a node and then click away from the node to close the context menu, and then right click again on a node the list of commits is still open. That list should probably always be closed every time I right click on a node.
> 2. Its unintuitive to me that I can right click on any node and delete any tag from that node. The list of tags to delete (in my opinion) should either only contain tags for that specific node, or the user just right clicks on the tag icon itself and selects the 'delete' option there.

I fixed the first issue with a small change. It's the latest commit. 

I thought that the second issue would be an easy fix, but after looking into it, it would require a refactor of how the tag list items are instantiated.

Issue #125, opened 2019-06-11
Allow a user to rename or move any file from the file editor panel


jpstoldt:
Functionality implemented under #49 introduced the ability to move or rename files with changes (staged/unstaged files). This issue will add buttons the file editor panel and allow the user to rename or move any file in the repository.

colinwerner:
Closing this issue on behalf of JP.

PR #164, opened 2019-06-17
Feature 125 move or rename any file

jpstoldt:
This feature adds file rename and move buttons the file editor panel and allows a user to rename or move any file in the repository, as described in issue #125.

### DIFF
--- a/app/components/diffPanel/diff.panel.component.html
+++ b/app/components/diffPanel/diff.panel.component.html
@@ -31,7 +31,7 @@
           </div>
       </div>
       <div class="modal-footer">
-          <button type="button" class="btn btn-primary" id="rename-file-button" (click)="renameFileFromDiff()">Rename File</button>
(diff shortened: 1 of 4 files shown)


jpstoldt:
> Great work! I'm able to easily do rename and move from both file editor and the staging area.
> 
> I did find a small bug which should not show up in normal use. If I do either rename or move **from the File Editor only**, do not make any changes, then click on Rename File or Move File, the error message shows up as expected. However, Once I try to click rename/move on the same file in editor again, the modal does not show up again.
> 
> This is a minor issue as users aren't likely to use rename/move from editor then not change the name/location before clicking the Rename/Move File button. Otherwise, the feature works just as intended.

Thanks - I have made additional changes that should address the issue identified by resetting the file picker before next use.

colinwerner:
> Looks good to me, the only issue is after committing the files I see a "No files staged" even though the commit goes through. (This is happening on the Master branch as well, so it's likely not related to your change; however, it is still an issue at hand).

Confirmed that this issue is not related (tracked via #171).

PR #165, opened 2019-06-17
Bug 122 large repositories dont open

sterlinglaird:
The biggest issue with opening large repositories was the highest zoom level ("basic"). Because it doesn't seem to be a very useful zoom level, I got rid of it entirely and replaced it with the second zoom level ("abstract"). I decided to keep the name "basic" in the code as it makes more sense than "abstract". There are still some performance issues with large repositories, mostly due to inefficient/redundant programming by the original authors but it is not quite a bit faster.

I also replaced the spinner loading screen with a progress bar to better indicate that the repo is loading.


SiRumCz:
can you provide a link to a large public repo for the test purpose? I didn't run into any problem with my test repos.

sterlinglaird:
@SiRumCz I had this happen on this repo. Sometimes it would load, just very slowly, other times it would enter an infinite loop. It had something to do with how it traversed the node id's.

sterlinglaird:
@roweba Good catch. The error was just a naming conflict with an existing element. Fixed now.

Issue #153, opened 2019-06-13
show stash information


colinwerner:
The user should be able to see information about the stash (much like the diff panel?).

colinwerner:
Work started in: issue-153-show-stash-info

Issue #126, opened 2019-06-11
Allow user to apply and drop a stash from the stash list


mehrler12:
Sub issue of #34 

User should be able to apply a stash by:
- Select stash on stash list
- Select apply
- See the stash has been applied to the repo

As well as drop a stash by:
- Select stash on stash list
- Select drop
- See the stash is dropped from the list

Issue #34, opened 2019-05-21
Add stashing and reapplying functionality


omazhary:
Add support for stashing and reapplying changes, so that users can store their changes without committing and move on to another branch. The user should be able to view what changes are stashed on what branches

colinwerner:
Decomposed into:
#86: Allow a user to stash changes
#87: Allow a user to view a list of stashed changes
#88: Allow a user to pop from the list of stashed changes

colinwerner:
#106: allow a user to stash untracked files

colinwerner:
Removing this from Sprint 3, as we continue to decompose it.

PR #168, opened 2019-06-17
Issue 126 apply/drop stashes

mehrler12:
The UI here is still waiting on approval from James so this PR isn't final.

There are a few things added in this PR:

- First, as per the issue #126 stashes can now be applied and dropped

- As per issue #153 stash information can now be shown by clicking on a specific stash

- The UI for the stash panel was fully redesigned to be more visual, it supports drag and drop. You can drop the stash onto the graph to pop it or the designated areas to apply or drop it

- The new position for the stash list required a reformat of the apps footer, the footer should now nicely resize to smaller screen sizes. Also an issue where you could scroll away the commit message box if you had too many commands on the terminal is solved.

- I also left the drag and drop stuff easily extendable if we want to support future drag and drop operations in the future.


mehrler12:
> Glad the drag and drop stuff managed to work with the graph! The error message that I got to run is also very clear.
> stasherror
> 
> A small thing: the "Pop!" button is a little confusing for someone who doesn't understand Git. The arrow icon has "Apply Stash" as a hover text, and having "Pop!" to apply the top of the stash to the repo is a bit unclear if the user doesn't understand that Stash is a queue. Is there maybe a clearer text that could be used?
> 
> Lastly: The list is unreadable in the other themes due to it's colour. 😒 The themes are kind of a P.I.B. but thought I should mention this.

@hdauvic What do you think of hover text for the pop button? "Pop" is terminology from git so i'd like to keep that there, how about something like "Apply and Drop Top Stash"?

mehrler12:
This PR also adds the last of the functionality for #34

Issue #111, opened 2019-06-08
Add interactive Rebase UI options


nikola-rados:
Add a table that would show each commit and have options for what to do with that commit (squash, pick, reword) as a part of the Rebase UI.

This is a sub-issue of #35.

aeogunmuyiwa:
Currently I have included a commit list pop up on rebase click which shows the commits. on branch feature-110-rebase-ui-popup-fresh

aeogunmuyiwa:
the link below contains images of the feature https://slack-files.com/TJBDCRM6G-FKM8JENTE-e3e69305d0
https://slack-files.com/TJBDCRM6G-FKPSN8L8N-721b562780 
https://slack-files.com/TJBDCRM6G-FKMNMBK6U-1eb49a5180

PR #169, opened 2019-06-17
Feature 110 rebase ui popup fresh

aeogunmuyiwa:
Issue #111 Added a table that would show each commit and have options for what to do with that commit (squash, pick, reword) as a part of the Rebase UI.


Athenasko:
image
When opening the interactive popout, there is a weird scrollbar that appears. In addition, if you click out of the interactive popout on the right of scrollbar, it freezes the application.

aeogunmuyiwa:
@Athenasko the scroll bar is present because I made wanted the second modal to display fully. If not it would look like these 
Screen Shot 2019-06-17 at 2 42 51 PM
and you would have to scroll through the second modal

Athenasko:
Couldn't you close the previous modal though before opening the new one? Would that not fix that problem?

aeogunmuyiwa:
@Athenasko Thanks for the catch, It was a design choice, However , I have fixed the issue. You can pull the latest changes and retest it . Thanks

nikola-rados:
Looks really good so far @aeogunmuyiwa! One quick suggestion would be to keep the position of the buttons consistent.  You have `Rebase` and `Abort` and then `Cancel` and `Continue Rebase`... I think it would be better if the order was `Continue Rebase` then `Cancel`. 👍

aeogunmuyiwa:
Thanks @nikola-rados , I have switched the order and pushed the changes

aeogunmuyiwa:
Branch deleted.

Issue #110, opened 2019-06-08
Add Rebase Skeleton UI


nikola-rados:
Add the rebase button to the sidebar as well as open a much larger slider that would cover the graph panel. Inside this panel we would have some UI features that would not yet be functional.

This is a sub-issue of #35.

aeogunmuyiwa:
Adebayo Ogunmuyiwa [12:04 PM]
Hey, guys so I created a modal pop up. I’m just working on a uI for the feature so this code is not in the app but I was wondering is this more in the direction of what we’re looking for ? 

https://gsd-seng421.slack.com/files/UJRK7RENT/FKGRT96NS/image_from_ios.jpg

nikola-rados:
More or less that is what we are looking for.  What we wanted to do though was try and avoid using a pop up because in the future we will have to handle merge conflicts.  This is because in a merge conflict you have to make commits and edit files... and since the tools for those things are already on the app we thought we would try and use them somehow.  That being said, both @Athenasko and I spent around 6 hours yesterday trying to implement this, and it was very difficult and unsuccessful.  So maybe we can go ahead with the pop-up for now, and worry about the merging and such for the future.

That being said some points to mention:
- I think we can just display the current branch, and have a dropdown menu for the branch to rebase onto
- I don't know if we need the preserve merge stuff, but leave it in there for now
- instead of `close` make the button `abort`

It looks really good @aeogunmuyiwa 

@Athenasko how do you feel about the popup? Have we gone deep enough to figure the panel just isn't worth the work or should we continue to explore that option?

aeogunmuyiwa:
I am still creating some UI elements and this code is independent of Visual Git and has not been added to the code yet. If there are any more suggestions on things to include let me know. I will integrate with the branch and push it on late on today.

Athenasko:
@aeogunmuyiwa Nik is right, it does look good. 

@nikola-rados I do still feel our concerns about merge conflicts, editing commits files are still valid. Perhaps we haven't explored all the possible options for the panel, but maybe in our call we can think of ways to deal with the future concerns while still using the pop-up?

nikola-rados:
@Athenasko and I looked into the panel option for another couple of hours, it does not seem to be possible at this point.  That being said we are going to go ahead with a popup and work from there.

aeogunmuyiwa:
Currently I have included a commit list pop up on rebase click which shows the commits. on branch feature-110-rebase-ui-popup-fresh

[...]

Issue #112, opened 2019-06-08
Add basic Rebase functionality


nikola-rados:
Add some basic functionality to rebase.  Have the user choose the branch they want to rebase onto and perform the rebase only if there are no merge conflicts (otherwise just abort).

This is a sub-issue of #35.

aeogunmuyiwa:
<img width="1183" alt="Screen Shot 2019-06-20 at 11 04 24 AM" src="https://user-images.githubusercontent.com/25914907/59870890-34d95780-934b-11e9-8b76-3ab497cb8b15.png">
Fixed the onto drop down list , I had to adjust how the options tag was created and call the function in the div

Athenasko:
Merged into master.

PR #170, opened 2019-06-18
Fixed the issue that created exceptions it was caused by the method t…

aeogunmuyiwa:
Fixed the issue that created exceptions it was caused by the method that called the get current branch, the code was updated in another issue and I just changed it here. #110 .


aeogunmuyiwa:
The issue originated from a change in outdated code, issue #112 . It called an ngFor function that caused the errors.

aeogunmuyiwa:
merging into master !!

Issue #187, opened 2019-06-19
layout of panels need to be fixed


Unimax:
![image](https://user-images.githubusercontent.com/4117562/59805948-9d6ef880-92a7-11e9-85d2-0300537734f5.png)

there is a gap in the panels which should be removed.

Unimax:
fixed with #190 PR.

Issue #181, opened 2019-06-19
Undo the latest unpush commit.


Unimax:
### Feature Addition

we should be able to undo a local only commit. that is a commit made in local repo and has not pushed to the remote.

Unimax:
fixed with #190 PR.

Issue #180, opened 2019-06-19
Check boxes on files can be aligned to text to improve UI


Unimax:
![image](https://user-images.githubusercontent.com/4117562/59804812-46b3ef80-92a4-11e9-9482-4c5ebcfad9ba.png)

the location of checkbox could be improved.

Unimax:
fixed with #190 PR.

Issue #179, opened 2019-06-19
Large scrollbar in case of too many changes


Unimax:
![before1](https://user-images.githubusercontent.com/4117562/59812049-53930c00-92c1-11e9-95ca-a73a9b58be9f.JPG)

if there are too many file changes we are not able to see area to stage files.

Unimax:
fixed with #190 PR.

PR #190, opened 2019-06-20
UI fixes for issue #171 ,#179 ,#180 ,#187  and feature to undo latest unpush commit.

Unimax:
This PR fixes:- 
#171 correct message on successful commit
#179 if we have too many files in the file-changed section it was a mess. fixed!
#180 checkbox for the files are now infront of them aligned to text. Note: file block size is still large to look prity. (i checked small does not look that cool and will create an uneven block in case of the large filename)
#187 as shown during planing the panels on both sides of the graph were little shorter then needed.fixed!

also #181 , you can now right-click an unpushed commit to undo the latest commit from that branch. 
Note: right-clicking on pushed commit does not have the undo option.

@jtiz003  demo for 181:- https://youtu.be/YXY0OlDC4ks

### DIFF
--- a/app/components/filePanel/file.panel.component.html
+++ b/app/components/filePanel/file.panel.component.html
@@ -1,10 +1,10 @@
 <div class="file-panel" id="file-panel">
   <!-- UNSTAGED FILES -->
   <div class="modified-files-header" id="modified-files-header">
-    <h4 style="color: white" id="unstaged-files-heading">Unstaged Files</h4>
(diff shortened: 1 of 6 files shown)


aeogunmuyiwa:
#171 works as expected 
#180 works as expected 
#187 Fixed now 
#181  works as expected 

Good job @Unimax  I have tested it and they all work good . Thanks :)

Unimax:
> Looks great! There is still a gap in the diff panel though, unsure if this is intentional
> image

fixed. thanks that was not part of the issue but i guess should be done.

PR #192, opened 2019-06-20
Ensured there is a commit to center on before centering.

sterlinglaird:
Really basic fix. In a new repo there are no commits so when loading a graph we need to test that there is a commit before trying to center on the first commit.

### DIFF
--- a/app/misc/graphing.ts
+++ b/app/misc/graphing.ts
@@ -695,7 +695,10 @@ function reCenter() {
         }
     };
 
-    network.focus(commitList[commitList.length - 1]["id"], moveOptions);
+    // Recenter on the first commit
+    if (commitList.length > 0) {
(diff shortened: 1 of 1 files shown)


sterlinglaird:
@dragonprevost This is the fix which will allow you to look at #5.

aeogunmuyiwa:
Looks good to me, the fix works now.

app/misc/graphing.ts
```diff
@@ -695,7 +695,10 @@ function reCenter() {
         }
     };
 
-    network.focus(commitList[commitList.length - 1]["id"], moveOptions);
+    // Recenter on the first commit
```
colinwerner: It's not really the first commit? It's actually the last commit?

Issue #178, opened 2019-06-19
Rework colours for unpushed commits


trstndvll:
Change the colours for the unpushed commit highlight to visually relate to the ahead behind badges on the right hand side. So, the node highlight should be the same colour of the ahead badge, and the pull button hover colour should be the same colour as the behind badge.

aeogunmuyiwa:
<img width="1183" alt="Screen Shot 2019-06-19 at 9 33 28 PM" src="https://user-images.githubusercontent.com/25914907/59818837-577f5800-92da-11e9-857a-93e2971a4400.png">

- Currently the ahead commit badge matched the ahead highlight of the graph

aeogunmuyiwa:
PR approved and merged into master

PR #193, opened 2019-06-21
Feature 178 Update Unpushed Commit Highlight Colours

trstndvll:
Issue #178  This PR changes the colors of the Ahead and Behind badges to match the highlight colors of the unpushed commit nodes in the graph. Ahead is colored red to match the highlight color, and Behind is colored blue to contrast the red (it isn't currently used for anything other than a reference).

Also added a feature where, on hover, the cursor changes to a specified icon. So for hovering on Ahead, the cursor will change to the push icon to tell the user they may want to push. Conversely, hovering on Behind will change the cursor to the pull icon.

### DIFF
--- a/app/components/pullRequestPanel/pull.request.panel.component.html
+++ b/app/components/pullRequestPanel/pull.request.panel.component.html
@@ -23,10 +23,12 @@
     <div class="right-pr-nav">
         <i (click)="togglePRPanel()"><img src="./assets/PullRequestIcon.svg"></i>
         <hr>
-        <a class="badge badge-pill badge-light text-left "  style="transform: scale(0.7);">Ahead
+        <!-- This badge's colour scheme matches the push icon's hover colour  (fa-cloud-upload:hover) -->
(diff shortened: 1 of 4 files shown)


trstndvll:
> May I request making it larger to match the text size of the things around it? Or is that because the numbers won't show on screen if that is done?

@hdauvic totally fair point. The only issue we run into is because we're using the badge element from bootstrap, it has very defined behaviour, one of which is clipping the sides of the div no matter what css properties we change (lol). I've been fighting with it for a while, but I can see if @aeogunmuyiwa or I have some time to try and fiddle with it a bit more. We'll let you know!

mehrler12:
> one of which is clipping the sides of the div no matter what css properties we change (lol)

It's a bit hacky but perhaps adding a negative left margin would work here?

aeogunmuyiwa:
@mehrler12 Yeah you're right, it is a bit hacky so we will just leave it the way it is and merge. both buttons are aligned and do their jobs perfectly.  Thanks for the review :)

aeogunmuyiwa:
Issue closed and merged into master

PR #194, opened 2019-06-21
only show the repo thing and the navbar when there is a repo path

colinwerner:
There were a couple ways to go about doing this, this was the simpliest.

Basically, only show the repo thing and the navbar when there is a repo path.

### DIFF
--- a/app/misc/router.ts
+++ b/app/misc/router.ts
@@ -19,7 +19,7 @@ function switchToClonePanel() {
   hidePullRequestPanel();
   hideGraphPanel();
   hideFooter();
-  hideNavToolbar();
+  checkRepoOpen();
   displayClonePanel();
 }
 
@@ -29,7 +29,7 @@ function switchToMainPanel() {
   displayFilePanel();
   displayPullRequestPanel();
   displayFooter();
-  displayNavToolbar();
+  checkRepoOpen();
   displayGraphPanel();
 
   openDisabled = false;
@@ -78,7 +78,7 @@ function switchToAddRepositoryPanel() {
   inTheApp = true
   console.log("Switching to add repo panel");
   useRecentRepositories();
-  hideNavToolbar();
+  checkRepoOpen();
   hideFooter();
   hideAuthenticatePanel();
   hideFilePanel();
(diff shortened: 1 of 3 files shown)


app/misc/router.ts
```diff
@@ -311,6 +314,22 @@ function hideDiffPanelButtons() {
...
+    document.getElementById("nav-repo-branch-tag-info")!.style.visibility = "hidden";
+    document.getElementById("nav-toolbar")!.style.visibility = "hidden";
+  }
+}
+
+function displayNavToolbar() {
+  // only show these if there is a repo
+  if (repoFullPath) {
+    document.getElementById("nav-repo-branch-tag-info")!.style.visibility = "visible";
+    document.getElementById("nav-toolbar")!.style.visibility = "visible";
```
roweba: Maybe you should try using the display property?

`document.getElementById("nav-repo-branch-tag-info")!.style.display = "none";` to hide
`document.getElementById("nav-repo-branch-tag-info")!.style.display = "block";` to show

maxUser:
The back button allows the user to access the graph page without having a repo open. Maybe remove it until a repo has been opened?

image

app/misc/router.ts
colinwerner: I don't think it's an issue with the method. I think it's an issue with the repoFullPath?

colinwerner:
> The back button allows the user to access the graph page without having a repo open. Maybe remove it until a repo has been opened?
> 
> image

Good catch. I'm not sure the best way to see if there is a repo open or not, any ideas?

colinwerner:
> Works on windows, the icon next to the repo navbar is still always visible but I think that's intentional and looks nice

Yep, that was intentional.

app/misc/router.ts
```diff
@@ -311,6 +314,31 @@ function hideDiffPanelButtons() {
...
+      repoCreate.value || repoClone.value || repoLocalPath)
+  {
+    showRepoNavTools = "visible";
+  }
+
+  // show/hide the relevent items
+  document.getElementById("repo-back-button")!.style.visibility = showRepoNavTools;
+  document.getElementById("nav-repo-branch-tag-info")!.style.visibility = showRepoNavTools;
+  document.getElementById("nav-toolbar")!.style.visibility = showRepoNavTools;
+  document.getElementById("nav-open-repo-button")!.style.visibility = showRepoNavTools;
```
mehrler12: Why'd you hide the folder button? I liked it being always visible

colinwerner: You and @maxUser can duke it out... let me know the result.

hdauvic: I concur with @mehrler12 but in addition to this if I do not have a repo open, there's no clear way of returning to the "open a repo" screen without the folder button. And as of right now (I know there's an issue to fix this) but visualgit always opens to that screen so now I can't open a repo at all.

colinwerner: > I concur with @mehrler12 but in addition to this if I do not have a repo open, there's no clear way of returning to the "open a repo" screen without the folder button. And as of right now (I know there's an issue to fix this) but visualgit always opens to that screen so now I can't open a repo at all.
> This is the screen visualgit opens to, and I can do nothing now as I have no repo open and no way to open one?
> repofolder

We either have to abandon this whole change or live with it. Why does VG open to this screen is the question?

roweba: Tested it with the new updates and I'm having the same problem as @hdauvic. VisualGit opens to the graph page by default, so I have no way of getting to the "open repo" page. I think you should either add the folder back or find a way to open VisualGit to the correct page.

colinwerner:
@maxUser you haven been outvoted ... for now.

@mehrler12 @roweba @hdauvic  please re-test now.

app/misc/router.ts
colinwerner: "Please try again." :)

Issue #176, opened 2019-06-19
Review overall codebase for error reporting consistency


hdauvic:
Currently, the code contains console line prints to aid with debugging that refer to specific portions/lines of code. With the mass of additions to the project, many of these references are now outdated, and incorrect.

**Steps to reproduce:**
This issue is difficult to reproduce as console logs only occur when very specific errors occur.

_Example_
Line 307 in git.ts:  `console.log("git.ts, line 203, cannot load all commits" + err);`
This line now points to a variable initialization, opposed to a function that may throw this specific error.

**Possible solutions**
1. Repoint all specific line references to where they should point, _or_
2. Remove line specifics, and just print the error.

PR #196, opened 2019-06-21
Editing errors for consistency

hdauvic:
This PR solves #176 , the specified line codes seem to have been a legacy issue and long out of date. The hard coded line numbers have been removed, and the error is printed to the debugging console - it will be up to future developers to trace the issue themselves if that error is ever thrown. 

This PR also solves #4 as I found the culprits for the repeated printing of the unnecessary log and have commented them out with an additional comment above them if future devs want to use those for debugging.

I also commented out the repetitive `url of repo` log, this console line was overwhelming at times if the user has copious amounts of repos associated with their username. Again, a comment was added to inform future dev's of this choice. While this solves the pictured issue in #20 it does **NOT** resolve that issue - with all the console logs of the graph, that feature is still highly recommended.

Finally, I combed through the console messages to update them for grammatical consistency, punctuation, clarity, and standardized error printing to include a descriptive error comment before printing out the error log. Each error now reads `ERROR` in capital letters for easier viewing in the debugging console when an error is thrown.

When reviewing this PR please ensure that none of the functionality that you've implemented has been broken by the addition of console logs. I doubt it would be, but I just want to make sure.

### DIFF
--- a/app/components/app/app.component.ts
+++ b/app/components/app/app.component.ts
@@ -24,7 +24,7 @@ export class AppComponent {
     // If user has previously saved a color, then set the app to that color
     if (fs.existsSync(userColorFilePath)) {
       fs.readFile(userColorFilePath, function(err, buffer) {
-        console.log(buffer.toString());
+        console.log("Last used theme: " + buffer.toString() + ", initializing to this theme.");
         let color = buffer.toString();
         changeColor(color);
       });
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -182,7 +182,6 @@ function authenticateUser(callback) {
     }, err => {
       console.log('ERROR while getting token. ', err);
 	}).catch( err => {
-    let tokenError = err;
     console.log("Token error: " + err);
   });
 }
--- a/app/misc/authenticate.ts
+++ b/app/misc/authenticate.ts
@@ -129,7 +129,7 @@ function searchRepoName() {
     for (let i = 0; i < data.length; i++) {
 
       let rep = Object.values(data)[i];
-      console.log("url of repo: " + rep['html_url']);
+      // console.log("url of repo: " + rep['html_url']);
 
       // Searches from the text input and adds to the list if repo name is found
       if (parseInt(rep['forks_count']) == 0) {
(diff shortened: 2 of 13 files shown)


app/misc/git.ts
```diff
@@ -39,7 +39,7 @@ function sortedListOfCommits(commits){
 
     while (commits.length > 0) {
       let commit = commits.shift();
-      let parents = commit.parents();
+      let parents = commit.parentxs();
```
Athenasko: Hi @hdauvic, would you mind elaborating on the change here? Was the addition of the 'x' a typo or is it a call in typescript? I couldn't find any reference and definitions other than here so I thought I'd ask for clarification. Thanks!

hdauvic: I'm seeing these x's pop up from when I merged the master to update mine. I didn't add this x, nor another that I randomly found in the code. The only code I touched was anything with "console.log" in front of it. This is worrysome.

hdauvic: I looked at the current master and have removed this x.

app/components/app/app.component.ts
```diff
@@ -24,7 +24,7 @@ export class AppComponent {
     // If user has previously saved a color, then set the app to that color
     if (fs.existsSync(userColorFilePath)) {
       fs.readFile(userColorFilePath, function(err, buffer) {
-        console.log(buffer.toString());
+        console.log("Last used theme: " + buffer.toString() + ", initializing to this theme.");
```
dragonprevost: I wonder if using the console log to notify theme change is really necessary? I feel like the console is a little crowded, but that is subjective. Let me know what you think...

app/misc/authenticate.ts
```diff
@@ -180,9 +180,10 @@ function authenticateUser(callback) {
       });
 
     }, err => {
-      console.log('Error while getting token', err);
+      console.log('ERROR while getting token. ', err);
 	}).catch( err => {
-    console.log(err);
+    let tokenError = err;
```
dragonprevost: Is `tokenError` used other than initializing it to equal `err` ?

dragonprevost:
Really the only thing that should maybe be changed is the tokenError variable. Great work other wise!

app/misc/authenticate.ts
hdauvic: Nope. Realized after the fact that I could just use err and it wouldn't cause any problems. Just didn't go back and fix it due to laziness.

Issue #186, opened 2019-06-19
Drag and drop for commiting


colinwerner:
### Feature Addition
Adding drag and drop functionality to committing an issue

mehrler12:
As part of this issue drag and drop guidelines were added for future features [here](https://github.com/uvicgse/project-virtual-team-1/wiki/Drag-and-drop-procedure)

Issue #171, opened 2019-06-18
Wrong message in popup when we do successful commit.


Unimax:
On successful commit, we are getting the wrong message in the popup.

Steps:-
1. make a change.
2. dont stage any file press commit.
3. press ok in the popup.
4. then stage files and commit.

Note: it seems like it showing whatever the previous popup text was.

Unimax:
fixed with #190 PR.

PR #199, opened 2019-06-21
Feature 186 drag/drop to commit

mehrler12:
This PR adds a few things for #186.

- Drag and drop to commit! You can now commit by dragging the commit box onto the graph

- Keyboard shortcut to commit, for the power users out there ctrl+enter will also activate the commit process.

- Commit button redesign, since the other 2 buttons are now gone the commit button looked weird on it's own

- Commit panel now disables when there are no staged files, this gives a better visual indicator of when you can and can't commit

- Lastly window.alert was causing issues with the drag and drop and was not in line with the rest of the project, this was brought in line with the rest of the project by using displaymodal instead

### DIFF
--- a/app/components/footer/footer.component.html
+++ b/app/components/footer/footer.component.html
@@ -1,8 +1,8 @@
 <div class="footer" id="footer">
     <div class="container-fluid" style="width: 100%;">
         <div class="row">
-            <div class="commit-panel col-sm-3" id="commit-panel">
-                <textarea placeholder="Commit disabled...&#10;&#10;No Files Staged" disabled rows=3 class="form-control" id="commit-message-input" draggable="true" onkeypress="commitShortcut(event)" ondragstart="handleCommitBoxDragStart(event)" ondragend="handleCommitBoxDragEnd(event)"></textarea>
+            <div class="commit-panel col-sm-3" id="commit-panel" draggable="false" ondragstart="handleCommitBoxDragStart(event)" ondragend="handleCommitBoxDragEnd(event)">
+                <textarea placeholder="Commit disabled...&#10;&#10;No Files Staged" disabled rows=3 class="form-control" id="commit-message-input"  onkeypress="commitShortcut(event)"></textarea>
                 <button class="input-group-addon commit-button-disabled" id="commit-button" onclick="addAndCommit()" disabled>Commit</button>
             </div>
 
--- a/app/components/footer/footer.component.html
+++ b/app/components/footer/footer.component.html
@@ -2,7 +2,7 @@
     <div class="container-fluid" style="width: 100%;">
         <div class="row">
             <div class="commit-panel col-sm-3" id="commit-panel">
-                <textarea placeholder="Describe your changes here..." class="commit-message-input" id="commit-message-input"></textarea>
+                <textarea placeholder="Describe your changes here...&#10;(Ctrl\Cmd + Enter or drag and drop to commit)" class="commit-message-input" id="commit-message-input" draggable="true" onkeypress="commitShortcut(event)" ondragstart="handleCommitBoxDragStart(event)" ondragend="handleCommitBoxDragEnd(event)"></textarea>
                 <div class="commit-button-wrapper" id="commit-button-wrapper">
                     <button class="commit-button" id="commit-button">Commit</button>
                     <button class="commit-button" id="fileEdit-button" (click)="displayFileEditor()">File Editor</button>
--- a/app/components/footer/footer.component.html
+++ b/app/components/footer/footer.component.html
@@ -2,7 +2,8 @@
     <div class="container-fluid" style="width: 100%;">
         <div class="row">
             <div class="commit-panel col-sm-3" id="commit-panel">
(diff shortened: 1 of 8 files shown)


subahm:
When I drag and dropped the commit box while the commit was disabled, it showed the message "Cannot commit without a message.....". I think we should update the message to something like "There is nothing to commit" for this scenario.

subahm:
I get the "cannot commit without a commit message...." even after I enter a message before committing, but the functionality works fine.

mehrler12:
@Athenasko 
> Is there any way to fix up the drag and drop such that it doesn't fade out weirdly?

From what I've read about drag and drop it doesn't seem like it, this comes from how chromium handles drag events.

> In addition, should the actual commit button be able to be dragged too since that would be more likely to be attempted than the text box?

Fixed! Thanks


> When you expand the commit box, it covers the command line text rather than pushing it over, and once its at the max size, it is no longer re-sizable and cannot be shrunk no matter what.

Didn't realize you could even do this! Disabled that feature, should no longer be a problem.

> In addition, the commit button without resizing hangs over the left panel side. Doesn't have to be fixed, but just wondering if was designed like that or if its resizing weirdly due to my screen size.

This was done when the footer was converted to use bootstrap. It makes the footer responsive but unfortunately the panel above it doesn't use bootstrap yet.   

> Lastly, the text box when typing throws a lot of keyboard events into the console logs. Is there anyway to suppress this or should those kinds of events always be record? (I'm actually not sure).

Whoops, left a debug log

> Is it possible to also make the commit button show the 'disabled' icon when hovering over it?

Done!

@subahm 

> When I drag and dropped the commit box while the commit was disabled, it showed the message "Cannot commit without a message.....". I think we should update the message to something like "There is nothing to commit" for this scenario.

Disabled the drag when committing is disabled

> I get the "cannot commit without a commit message...." even after I enter a message before committing, but the functionality works fine.

I think this is #171

Athenasko:
@mehrler12 
> From what I've read about drag and drop it doesn't seem like it, this comes from how chromium handles drag events.

How does the commit drag and drop differ from the staged/unstaged changes and stash drag and drops? Those don't seem to have the same problems.

> Didn't realize you could even do this! Disabled that feature, should no longer be a problem.

Great thanks! XD

> This was done when the footer was converted to use bootstrap. It makes the footer responsive but unfortunately the panel above it doesn't use bootstrap yet.

Sounds good. Thanks for clarifying!

mehrler12:
> How does the commit drag and drop differ from the staged/unstaged changes and stash drag and drops? Those don't seem to have the same problems.

Can you send me a screen shot? They do that for me

Athenasko:
> Can you send me a screen shot? They do that for me. Perhaps cause the element is larger the fade out is more pronounced?

That's really odd. I just checked it out and they are doing that now. They shouldn't be though. Previous behaviour it showed the entire section being dragged as it should be. 
unstagedCommitsBefore

I'm not sure what changed, but all the drag and drop stuff should be acting as the image above and not fading.

trstndvll:
Screenshot 2019-06-22 14 25 47

Perhaps this is the cause? May or may not be related, but in header.component.html, starting at line 467, `<div id="delete-upstream-modal" class="modal fade">` seems to have 4 open divs that don't get closed.

mehrler12:
@Athenasko 

> I'm not sure what changed, but all the drag and drop stuff should be acting as the image above and not fading.

So I've been fiddling with this a bunch and this doesn't seem to be something I can affect at all, I can set a new drag image but this also is faded out similarly to the div. I could totally turn off the ghost image if you think that would be better but short of writing our own drag and drop api I don't think we can fix this

@trstndvll 

> Perhaps this is the cause? May or may not be related, but in header.component.html, starting at line 467, `<div id="delete-upstream-modal" class="modal fade">` seems to have 4 open divs that don't get closed.

Totally unrelated but might as well sneak in a fix

Athenasko:
> So I've been fiddling with this a bunch and this doesn't seem to be something I can affect at all, I can set a new drag image but this also is faded out similarly to the div. I could totally turn off the ghost image if you think that would be better but short of writing our own drag and drop api I don't think we can fix this

Huh that's too bad. It might just be better to turn off the ghost image I think. Odd that it changed though. Thanks for investigating it.

Athenasko:
Really sorry @mehrler12. I definitely misunderstood what turning off the ghost image meant. In my head it meant showing the fully colored draggable item just without the fade effect as previous drag and drops have had, and without the ghost effects that is happening now. I don't know exactly why I thought that, but since this is a visual app, if it's not too much trouble I think the ghost effect would be more effective than without it since it becomes confusing.

mehrler12:
> Really sorry @mehrler12. I definitely misunderstood what turning off the ghost image meant. In my head it meant showing the fully colored draggable item just without the fade effect as previous drag and drops have had, and without the ghost effects that is happening now. I don't know exactly why I thought that, but since this is a visual app, if it's not too much trouble I think the ghost effect would be more effective than without it since it becomes confusing.

Yeah, disabling it also breaks a bit of functionality on Mac. I disabled it to check it with James but yeah I think having it is better then not! Thanks for the comments @Athenasko !

PR #203, opened 2019-06-23
Issue #61 tag checkout feature and few minor fixes

SiRumCz:
Feature:
- allow user to checkout any tag

Fixes:
- fixed problem with growing '...' in repo name in navigation bar
- put require() all on the top
- fixed graphing duplicate node error

Description:
original idea was to show 'Tag: xxxxx' in the navigation bar to indicate a tag has been chackout out. After I looked into NodeGit API, the reference it returned is an isolated branch named "HEAD"which points to the same commit as checked out tag. I then decided to use "HEAD detached at xxx" just the same as it showed in the demo below. and a new branch HEAD will be created which also points to the same commit node as the tag that has been checkout.

Screen shots/Demo:
image
image
image
image


sterlinglaird (approved):
Looks good. I'm sure its fine that it doesn't say which tag is checked out as this is the last sprint and is supposed to be a lighter work load.

hdauvic (approved):
Works well, but I was confused as to _how_ to checkout a tag - but I'm not sure there's an easy way to indicate this without massive text, plus there's no indication to click to checkout a branch either.

maxUser (approved):
Works as intended!

Issue #201, opened 2019-06-23
Fix "open in editor" button in Diff panel


jpstoldt:
Pressing "Open in Editor" from Diff panel does not open the text editor.

Issue #159, opened 2019-06-14
If diff panel is open, you can't save a file using file editor


hdauvic:
When you click on a changed file in the staged file section to view the diff panel, then proceed to change said file, you cannot save the changes.

**Steps to Reproduce**
1. Open file editor
2. Choose a file to edit
3. Edit a file
4. Save changes, and close
5. Click the file in the unstaged files section
![diff1](https://user-images.githubusercontent.com/31779196/59543396-e138bb00-8ebf-11e9-844b-9d1a46d5fc16.png)
6. Click `open in editor`
![diff2](https://user-images.githubusercontent.com/31779196/59543406-ebf35000-8ebf-11e9-86df-7875e30ab68c.png)
7. Edit the file
8. Click Save
![diff3](https://user-images.githubusercontent.com/31779196/59543422-075e5b00-8ec0-11e9-8806-0f024bce91e5.png)

**Current Behaviour**
The * will not disappear from beside the file's name, and save will not work. If you close the editor, close the diff panel by clicking the file again, and reopen the diff panel the second changes will not appear.

PR #204, opened 2019-06-24
Fix "Open in Editor" button in Diff Panel

ghost:
This PR fixes issue #201. Pressing "Open in Editor" in Diff panel now opens the text editor for selected file.

### DIFF
--- a/app/components/diffPanel/diff.panel.component.ts
+++ b/app/components/diffPanel/diff.panel.component.ts
@@ -1,5 +1,5 @@
 import { Component } from "@angular/core";
-import { FooterComponent } from "../footer/footer.component";
+import { HeaderComponent } from "../header/header.component";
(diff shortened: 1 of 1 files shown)


colinwerner (approved):
So the issue is/was that you moved this component to the header as opposed to the footer?

jpstoldt:
> So the issue is/was that you moved this component to the header as opposed to the footer?

There was a broken object reference since displayFileEditor() is a method of the HeaderComponent class and not FooterComponent

maxUser (approved):
Good stuff!

aeogunmuyiwa:
looks good and works!

aeogunmuyiwa (approved):

hdauvic (approved):
Works as intended, but it should be noted that this does _not_ fix issue #159 - just in case.

PR #205, opened 2019-06-24
Issue 175 commit functionality from navbar

subahm:
- User should be able to commit from navbar

-Clicking on the commit button should open a commit modal

-User should be able to enter a message and commit their changes

### DIFF
--- a/app/components/header/header.component.html
+++ b/app/components/header/header.component.html
@@ -465,6 +465,27 @@
   </div>
 </div>
 
+<!--Modal for commiting from navbar-->
+<div id="set-commit-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="mySmallModalLabel"
+  aria-hidden="true">
+  <div class="modal-dialog modal-sm">
+    <div class="modal-content" style=" width: 602px !important">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+        </button>
+        <h4 class="modal-title">Commit changes</h4>
+      </div>
+      <div class="modal-body" id="modal-text-box">
+        Please provide a commit message.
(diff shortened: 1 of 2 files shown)


app/misc/git.ts
```diff
@@ -125,6 +125,9 @@ function stage() {
 
 function addAndCommit() {
   commitMessage = document.getElementById('commit-message-input').value;
+  if (commitMessage == null || commitMessage == "") {
+    commitMessage = document.getElementById('commit-message-input-navbar').value;
+  }
   if (commitMessage == null || commitMessage == "") {
     displayModal("Cannot commit without a commit message. Please add a commit message before committing");
     return;
```
jpstoldt: This sequence of if blocks would benefit from some comments explaining which commit message is used for the git commit and why.

app/components/header/header.component.html
```diff
@@ -465,6 +465,26 @@ <h5 type="text">Upstream:</h5>
...
+  aria-hidden="true">
+  <div class="modal-dialog modal-sm">
+    <div class="modal-content" style=" width: 602px !important">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+        <h4 class="modal-title">Commit changes</h4>
+      </div>
+      <div class="modal-body" id="modal-text-box">
+        Please provide a commit message.
+        <input type="text" id="commit-message-input-navbar" class="form-control" style=" width: 554px !important" placeholder="Enter a commit message">
```
jpstoldt: I am wondering if we should pre-populate this textbox with the text entered for a commit message on the main screen?

aeogunmuyiwa:
The changes work as described. Looks good to me

app/misc/git.ts
subahm: Good idea. I have added some comments to explain the sequence of if statements and getting value of elements from 2 different ids

PR #206, opened 2019-06-24
Issue 172 - Edit Upstream Repo

### LINKED ISSUE #172: Edit/Delete the Upstream in One Command
Currently when editing the upstream repository, the user has to enter the url of the new upstream and then click confirm. However, if an upstream is already configured then an error is thrown. The user has to manually delete the upstream themselves before adding a new one. A user should be able to enter the address of the new upstream, and then a modal should appear that says "Upstream already exists, do you want to replace with {url}?" and the user will cancel or confirm.

roweba:
Fixes #172: Previously, in order to edit the upstream repo you had to manually delete the upstream before entering the url of the new upstream. Now, if you have an upstream configured and enter a new one, a confirmation modal will appear and ask if you want to edit the upstream.

### DIFF
--- a/app/misc/git.ts
+++ b/app/misc/git.ts
@@ -1772,6 +1772,7 @@ function editUpstream() {
     Git.Repository.open(repoFullPath)
       .then(function (repo) {
       repository = repo;
+      addCommand("git remote rm upstream");
       Git.Remote.delete(repository, 'upstream').then(function(result) {
         addCommand("git remote add upstream " + upstreamRepoPath);
         Git.Remote.createWithFetchspec(repository, 'upstream', upstreamRepoPath, '+refs/heads/*:refs/remotes/upstream/*').then(function(remote) {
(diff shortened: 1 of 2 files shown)


roweba:
@colinwerner updated!!

app/misc/git.ts
```diff
@@ -1773,6 +1773,7 @@ function editUpstream() {
       .then(function (repo) {
       repository = repo;
       Git.Remote.delete(repository, 'upstream').then(function(result) {
+        addCommand("git remote add upstream " + upstreamRepoPath);
```
colinwerner: I think it should do a delete first -- then the add?

roweba: good catch, it's updated now

colinwerner:
Forgot my comment, but thank you for the quick changes to my address my comments, Rowena. Ship it!

Issue #157, opened 2019-06-14
GitHub profile pictures on nodes, for commit authors with non-noreply emails


norrisng:
This is a sub-task of #31.

PR #135 partially implements the above feature; in particular, it only works when the commit author has a `@users.noreply.github.com` email.

This issue would expand this functionality to all commits, regardless of the commit author's email.

https://github.com/uvicgse/project-virtual-team-1/pull/135#pullrequestreview-248926557 mentions that this can be achieved using the `github-username` package.

norrisng:
See #207 (branch `issue-157-profile-pictures-all`) for partial implementation and associated notes. 

c6c3c813487af320bf4197dbbecb5de38d187d4f re-implements profile picture retrieval entirely by looking up the commit hash's author on the GitHub API (instead of using the email in any form), but this will still run into rate limits. again due to the way the graph nodes are added asynchronously.

There is no API endpoint that provides all of a repository's commits, but there is one (`/repos/{owner}/{repository_name}/contributors`) that provides a list of all contributors (but without their email).

PR #207, opened 2019-06-24
[WIP] Issue 157: GitHub profile pictures for all nodes

norrisng:
This PR implements #157. Well, that's what it _attempts_ to do anyway.

However, it currently fails for anything beyond a modestly-sized repo, as the rate at which the nodes are being added (and consequently the number of API calls) will exceed the GitHub API rate limit. 

While `images.ts` _does_ cache usernames/avatar URLs, VisualGit will continue to add nodes (and make API calls) without waiting for the API calls to return, so the cache doesn't get a chance to be populated beyond the first commit's user. By the time the second commit's API call returns, too many API calls will have been made since, triggering the rate limit.

I'll continue trying to get this to work for the rest of the day, but at this point it's looking like it requires significantly more work than I originally anticipated.

### DIFF
--- a/app/misc/graphing.ts
+++ b/app/misc/graphing.ts
@@ -546,19 +546,26 @@ function makeNode(c, column: number, isUnpushCommit : boolean) {
         }
     }
 
-    nodes.add({
-        id: id,
-        shape: "circularImage",
-        title: title,
-        image: img4User(name),
-        physics: false,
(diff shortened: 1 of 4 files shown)


norrisng:
See https://github.com/uvicgse/project-virtual-team-1/issues/157#issuecomment-505991643

PR #208, opened 2019-06-26
Basic Rebasing Functionality

nikola-rados:
We have implemented the basic functionality for `git rebase`.  Dropdown menus have been populated with both local and remote branches which can be selected for the rebase.  The implementation was done using `simple-git` since the `nodegit` alternative was not working correctly for us.

### DIFF
--- a/app/components/header/header.component.html
+++ b/app/components/header/header.component.html
@@ -523,146 +523,124 @@
   </div>
 </div>
 
-<!--rebase-->
-<div id="rebase-modal" class="modal fade  " tabindex="-1" role="dialog" aria-hidden="true">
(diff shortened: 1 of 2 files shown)


aeogunmuyiwa:
It tested a basic rebase with no merge conflict and it works. a note should be left somewhere for people carrying on the project to implement the close when It is done . but the functionality works. I approve.

Athenasko:
@trstndvll 
With how the Interactive UI is currently coded, getting the model to disappear actually breaks rebase if the interactive flag is set. Nik and I discussed, and in this case without having the time to properly get the interactive changes working, the normal one has to do without the automatic closing unfortunately.

